PR #72, opened 2015-04-13
Get mediaTypeRegistry with MediaTypeRegistry.getDefaultRegistry

jnioche:
and instanciate Tika from the start.


kkrugler:
I like the naming convention of all uppercase for static finals, so TIKA vs. tika and MEDIA_TYPE_REGISTRY versus mediaTypeRegistry.

Otherwise seems fine to me.

jnioche:
Good idea! Thanks Ken

PR #75, opened 2015-04-29
Sitemaps : more robust parsing of XML elements

jnioche:
This patch simplifies the logic of extracting textual values from elements such as `loc`. At the moment it throws a NPE if the first child element does not have a value as in the example below \: 

[http://jobs.optistaffing.com/sitemap.xml]

```
<url>
<loc>
<url>
<![CDATA[
http://jobs.optistaffing.com/EXPERIENCED-DISPATCHER-NEEDED-NOW----Jobs-in-Vancouver-WA/2333221
]]>
</url>
</loc>
<lastmod>2015-04-28</lastmod>
<changefreq>daily</changefreq>
</url>
```

at least we should handle this more gracefully and check that `children.item(0)).getNodeValue()` is not null.

This PR uses a more generic approach for extracting the textual content of a node which also has the added benefit of extracting correctly from the cases above where the content is slightly incorrect e.g. additional url element within the loc.

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
+++ b/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
@@ -570,10 +570,7 @@ public class SiteMapParser {
         NodeList list = elem.getElementsByTagName(elementName);
         Element e = (Element) list.item(0);
         if (e != null) {
-            NodeList children = e.getChildNodes();
-            if (children.item(0) != null) {
-                return ((Node) children.item(0)).getNodeValue().trim();
-            }
+            return e.getTextContent();
         }
 
         return null;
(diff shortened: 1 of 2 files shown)


jnioche:
@crawler-commons/owners - any objections to committing this?

kkrugler:
Hi Julien,

If we're trying to protect against bad data, then shouldn't getElementValue() check if getElementsByTagName() returns a non-null list with at least one element before continuing?

In general the code in SiteMapParser seems to assume the resulting DOM will contain well-formed data, e.g. line 536 is iterating over what could be a null result if the RSS text isn't valid. From my experience in Tika, consumers of parse results would rather have logging & no NPEs versus having to catch errors, especially since one bit of malformed text in a large sitemap would then render the entire map useless.

So wondering if we should have a separate issue for bullet-proofing the code.

jnioche:
Thank Ken. Definitely! No need for a separate issue, let's do that in this one.

jnioche:
Committed in [master 20861ba] Issue 75: [Sitemaps] more robust parsing of XML elements (jnioche, kkrugler)

jnioche:
new issue for this #79

PR #76, opened 2015-04-30
maven-java-formatter-plugin

### LINKED ISSUE #49: Proper code styling template and mechanism to enforce it
```
Ken has committed in the past an Eclipse formatting file:
http://code.google.com/p/crawler-commons/source/browse/trunk/doc/eclipse-formatt
er.properties

We should have a generic styling file which will match all IDEs (I personally 
use IntelliJ Idea for example).

If the styling template is accepted by us we should use it, or we can compare 
it to other styling files used by other groups and upgrade it if we see fit.


We should also find a mechanism which will use this file in Maven at the 
"Compile" task automatically.

Possible candidate will be:
http://code.google.com/p/maven-java-formatter-plugin/




Summary:
* Find suitable styling guidelines
* Enforce it using a generic (xml ?) file
* Find a good maven plugin which will run it every compile
```

Original issue reported on code.google.com by `avrah...@gmail.com` on 9 Aug 2014 at 6:58

jnioche:
Fixes #49 

The command `mvn java-formatter:format` will apply the formatting to the code using the Eclipse formatting provided.

The latter can be imported directly in Eclipse and there seems to be plugins for other editors, however the point here is that the formatting will be done regardless of which editor is used.

We can discuss / improve / modify the actual formatting choices but at least we'll have a mechanism for enforcing it.

Please review, thanks!


jnioche:
@crawler-commons/owners - any objections to committing this?

kkrugler:
Hi Julien - looks good, thanks!

Two things I noticed...

1. We now have doc/eclipse-formatter.properties and doc/eclipse-formatter.xml. I think you can get rid of the .properties version, as that was from when we were using ant (and the ant-eclipse task needed a .properties file to create the Eclipse project)

2. The eclipse-formatter.xml file you provided is for a 1.5-compliant project. E.g. compiler.source=1.5, etc. We're at 1.6 now, and I think if you switch to a 1.6 project, there will be extra settings. So I think using a 1.6-compliant formatting file would be a better starting point.

jnioche:
Thanks Ken

Issue #79, opened 2015-05-22
Sitemaps : more robust parsing of XML elements


jnioche:
See #75 

More work is needed on making the parser bullet-proof

jnioche:
Closing now that the SAX parsing has been merged

PR #80, opened 2015-05-27
Parse XML sitemaps via SAX for option to parse partial/broken documents

matt-deboer:
Adds an additional optional constructor arg (false by default) which enables parsing of partial/broken xml-based sitemap documents.

### DIFF
--- a/crawler-commons.iml
+++ b/crawler-commons.iml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module org.jetbrains.idea.maven.project.MavenProjectsManager.isMavenModule="true" type="JAVA_MODULE" version="4">
+  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_1_6" inherit-compiler-output="false">
+    <output url="file://$MODULE_DIR$/target/classes" />
+    <output-test url="file://$MODULE_DIR$/target/test-classes" />
+    <content url="file://$MODULE_DIR$">
+      <sourceFolder url="file://$MODULE_DIR$/src/main/java" isTestSource="false" />
+      <sourceFolder url="file://$MODULE_DIR$/src/main/resources" type="java-resource" />
+      <sourceFolder url="file://$MODULE_DIR$/src/test/java" isTestSource="true" />
+      <sourceFolder url="file://$MODULE_DIR$/src/test/resources" type="java-test-resource" />
+      <excludeFolder url="file://$MODULE_DIR$/target" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+    <orderEntry type="library" name="Maven: org.apache.httpcomponents:httpclient:4.3.5" level="project" />
+    <orderEntry type="library" name="Maven: org.apache.httpcomponents:httpcore:4.3.2" level="project" />
+    <orderEntry type="library" name="Maven: commons-logging:commons-logging:1.1.3" level="project" />
+    <orderEntry type="library" name="Maven: commons-codec:commons-codec:1.6" level="project" />
+    <orderEntry type="library" name="Maven: commons-io:commons-io:2.4" level="project" />
+    <orderEntry type="library" name="Maven: org.apache.tika:tika-core:1.8" level="project" />
+    <orderEntry type="library" name="Maven: org.slf4j:slf4j-api:1.7.7" level="project" />
+    <orderEntry type="library" scope="TEST" name="Maven: junit:junit:4.7" level="project" />
+    <orderEntry type="library" scope="TEST" name="Maven: org.slf4j:slf4j-log4j12:1.7.7" level="project" />
+    <orderEntry type="library" scope="TEST" name="Maven: log4j:log4j:1.2.17" level="project" />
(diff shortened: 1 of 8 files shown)


jnioche:
Thanks @matt-deboer, looks interesting. Do you expect any performance gains compared to the current DOM-based parser?
Could you please apply the formattting as per [https://github.com/crawler-commons/crawler-commons/wiki/Code-Formatting] and squash your commits? This will make it easier to review this PR.
Thanks

matt-deboer:
Sure, done.

In general, performance gains are expected, especially for very large sitemaps (haven't measured that aspect yet; I'll reply back with some quick measurements)--the real driver was to simplify the case where a broken sitemap was delivered (I have a particular focused crawl in mind); with DOM, I had to manually patch the contents and re-process them. Yeah, one could argue that broken sitemaps shouldn't exist, but not everyone you want/have to crawl is up to speed on their SEO : )

There could be more gains with the parsing of really large sitemaps if an additional public **parseSiteMap** method were added to take in an InputStream instead of byte[] -- allowing the caller to avoid pulling the whole contents into memory at once.

What do you think?

Chaiavi:
Is it needed optimization ?

As the limit for a sitemap is 10mb

matt-deboer:
This is not primarily a perf optimization--it is a "robustness" improvement
to allow parsing of broken site maps. Any perf improvement is an added
bonus...

kkrugler:
Hi Matt,

I think Avi was asking specifically for any additional details about:

> > > There could be more gains with the parsing of really large sitemaps if an
> > > additional public *parseSiteMap* method were added to take in an
> > > InputStream instead of byte[] -- allowing the caller to avoid pulling the
> > > whole contents into memory at once.

e.g. have you seen issues with having to load the sitemap into memory first.

But in any case, seems very useful to handle broken site maps, since as you note it's common to get busted files in the real world.

Also, if there were no regressions w/using a SAX-based approach, then what would be reasons for sticking with the existing DOM approach?

Thanks,

-- Ken


> From: Matt DeBoer
> Sent: May 28, 2015 9:44:18am PDT
> To: crawler-commons/crawler-commons
> Subject: Re: [crawler-commons] Parse XML sitemaps via SAX for option to parse partial/broken documents (#80)
> 
> This is not primarily a perf optimization--it is a "robustness" improvement
> to allow parsing of broken site maps. Any perf improvement is an added
> bonus...
> 
> On Thursday, May 28, 2015, Avi Hayun <notifications@github.com> wrote:
> 
> > Is it needed optimization ?
> >
> > As the limit for a sitemap is 10mb
> >
> > On Thu, May 28, 2015 at 3:00 AM, Matt DeBoer <notifications@github.com
> > <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> > wrote:
> >
> > > Sure, done.
> > >
> > > In general, performance gains are expected, especially for very large
> > > sitemaps (haven't measured that aspect yet; I'll reply back with some
> > quick
> > > measurements)--the real driver was to simplify the case where a broken
> > > sitemap was delivered (I have a particular focused crawl in mind); with
> > > DOM, I had to manually patch the contents and re-process them. Yeah, one
> > > could argue that broken sitemaps shouldn't exist, but not everyone you
> > > want/have to crawl is up to speed on their SEO : )
> > >
> > > There could be more gains with the parsing of really large sitemaps if an
> > > additional public *parseSiteMap* method were added to take in an
> > > InputStream instead of byte[] -- allowing the caller to avoid pulling the
> > > whole contents into memory at once.
> > >
> > > What do you think?
> > >
> > > —
> > > Reply to this email directly or view it on GitHub
> > > <
> > https://github.com/crawler-commons/crawler-commons/pull/80#issuecomment-106115526
> > >
> > > .
> > >
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/crawler-commons/crawler-commons/pull/80#issuecomment-106218177>
> > .
> >
> —
> Reply to this email directly or view it on GitHub.
> 

--------------------------
Ken Krugler
+1 530-210-6378
http://www.scaleunlimited.com
custom big data solutions & training
Hadoop, Cascading, Cassandra & Solr





--------------------------
Ken Krugler
+1 530-210-6378
http://www.scaleunlimited.com
custom big data solutions & training
Hadoop, Cascading, Cassandra & Solr

jnioche:
Would this solve #79?

matt-deboer:
Sorry for the extended delay; just got back to this tonight.

@Chaiavi, @kkrugler: 
My original guesses about potential gains and "large" sitemaps proved to be wrong. After some testing with 10MB sitemaps, the performance is roughly the same (by roughly, I mean averaged about the same after 10 runs of each on my macbook). In fact, the initial implementation using a Stack was actually worse, so I swapped it for a LinkedList. My earlier comment about adding an additional *parseSiteMap* method can be disregarded.
To sum up, this PR is all about parsing of potentially broken site maps. 

@jnioche: 
I added a test case that included your example referenced in #79 (from #75), and your example is parsed successfully in this new implementation.

lewismc:
Apart from this patch offering conflicts with master, is there a compelling reason not to update and merge folks?

jnioche:
Would it make sense to have that as well as the DOM-based parser for some time? Could for instance have a separate package for this

kkrugler:
Hi @jnioche - thanks for following up on this (old) issue...

PR #82, opened 2015-06-11
[Sitemaps] Upgrade Valid / Legal / Strict SitemapUrls

### LINKED ISSUE #60: [Sitemaps] Upgrade Valid / Legal / Strict SitemapUrls
```
SitemapUrls can be not valid when they are referenced in a sitemap which it's 
directory is on a completely different path than the referenced SitemapUrl.

All as indicated here:
http://www.sitemaps.org/protocol.html#location


In order to clarify the validity aspect we need to upgrade the following
1. Add a little more explanations as javadocs and as logs
2. Rename "Legal" (I think only one occurrence) to "valid" (in the parser)
3. Add to the Sitemap class a new method to get all *valid* SitemapUrls
4. When dropping a URL due to invalidity a log should be shown, a URL shouldn't 
be dropped quietly.
```

Original issue reported on code.google.com by `avrah...@gmail.com` on 9 Nov 2014 at 2:10
* Blocked on: #57

jnioche:
Fixes #60 

SitemapUrls can be not valid when they are referenced in a sitemap which
it's
directory is on a completely different path than the referenced
SitemapUrl.

All as indicated here:
http://www.sitemaps.org/protocol.html#location

In order to clarify the validity aspect we need to upgrade the following
1. Add a little more explanations as javadocs and as logs
2. Rename "Legal" (I think only one occurrence) to "valid" (in the
parser)
3. Add to the Sitemap class a new method to get all *valid* SitemapUrls
4. When dropping a URL due to invalidity a log should be shown, a URL
shouldn't
be dropped quietly.

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/SiteMap.java
+++ b/src/main/java/crawlercommons/sitemaps/SiteMap.java
@@ -30,7 +30,9 @@ public class SiteMap extends AbstractSiteMap {
      */
     private String baseUrl;
 
-    /** URLs found in this Sitemap */
+    /**
+     * URLs found in this Sitemap
+     */
     private List<SiteMapURL> urlList;
 
     public SiteMap() {
@@ -68,8 +70,7 @@ public class SiteMap extends AbstractSiteMap {
     }
 
     /**
-     * @param url
-     *            - the URL of the Sitemap
+     * @param url - the URL of the Sitemap
      */
     private void setUrl(URL url) {
         this.url = url;
@@ -77,9 +78,8 @@ public class SiteMap extends AbstractSiteMap {
     }
 
     /**
-     * @param url
-     *            - the URL of the Sitemap In the case of a Malformed URL the
-     *            URL member is set to null
+     * @param url - the URL of the Sitemap
+     *            In the case of a Malformed URL the URL member is set to null
      */
     private void setUrl(String url) {
         try {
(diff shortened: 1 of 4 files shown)


src/main/java/crawlercommons/sitemaps/SiteMap.java
```diff
@@ -94,11 +96,8 @@ private void setUrl(String url) {
     public String toString() {
         StringBuilder sb = new StringBuilder();
 
-        Date lastModified = getLastModified();
-        String lastModStr = (lastModified == null) ? "null" : SiteMap.getFullDateFormat().format(lastModified);
-
-        sb.append("url = \"").append(url).append("\", lastMod = ").append(lastModStr).append(", type = ").append(getType()).append(", processed = ").append(isProcessed()).append(", urlListSize = ")
-                        .append(urlList.size());
+        sb.append("url = \"").append(url).append("\", lastMod = ").append((getLastModified() == null) ? "null" : SiteMap.getFullDateFormat().format(getLastModified())).append(", type = ")
```
jnioche: The existing code is easier to read and debug. Inlining it does not add anything, does it?

lewismc: I would agree @jnioche. The one line concatenation is almost cryptic.

lewismc:
It would also be nice to have a test case which asserts the statement "A SitemapUrl(s) can be not valid when they are referenced in a sitemap with which it's directory is on a completely different path than the referenced SitemapUrl."
Any thoughts?

Chaiavi:
The cryptiness comes from our automatic styling plugin we run now in order
to have consistent code.

That ugly line, was styled like all other chained methods, where each
method was started on a different line as can be seen here:
https://github.com/crawler-commons/crawler-commons/commit/feb40af519ebe40a0a98d903778b0caaa2fa7a50

In the old style it made sense as I didn't see any added value in those two
temporary variables.


That being said, now I agree that it is less readable.


Will revert it.




A. Does it mean that all other code was reviewed ?
B. I think it is a good test case, we just need to have in mind the future
implementation of the cross submit feature

lewismc:
@Chaiavi 
 bq. A. Does it mean that all other code was reviewed ?
Yes
bq. I think it is a good test case, we just need to have in mind the future implementation of the cross submit feature
Agreed. 

@Chaiavi can you commit this to master?

Chaiavi:
Yes,  I will push it forward.

jnioche:
in Sitemap - why wasn't the line reverted as you said you would? Could you please do that in separate commit?
Thanks

Chaiavi:
Didn't forget it, i just hoped the "Merge pull request" on github will give
me more control (where i could do some manual code change) - I was wrong.

I will change it back in a seperate small commit.

PR #91, opened 2015-08-28
handling of robots.txt with multiple matching user-agent lines

MichaelRoeder:
Hi all,

first of all, thanks for sharing this project! :)

### Improvement

I encountered a problem with the parsing of `robots.txt` files. I saw examples in which a user-agent has been defined more than ones. An example for this is http://datahub.io/robots.txt containing `User-agent: *` and `User-Agent: *` in a second paragraph. From my point of view this is not really nice, but it does not seem to be prohibited.

### Problem

I adapted the code of the SimpleRobotRulesParser class. Unfortunately, one of the test cases is not green any more. From my point of view, this could be a problem of the test case. Thus, before merging this pull request we might discuss the `testNutchCases` method of the JUnit test.

The robot.txt file of the test case looks like this:
```
User-Agent: Agent1 #foo  
Disallow: /a  
Disallow: /b/a
#Disallow: /c  

User-Agent: Agent2 Agent3#foo  
User-Agent: Agent4
Disallow: /d  
Disallow: /e/d/  

User-Agent: *
Disallow: /foo/bar/
```
This file is stored in the `content` variable and parsed as follows:
```
SimpleRobotRulesParser robotParser = new SimpleRobotRulesParser();
return robotParser.parseContent("http://domain.com", content, "text/plain", "Agent5,Agent2,Agent1,Agent3,*");
```
The parser should look for rules for the agents `Agent5,Agent2,Agent1,Agent3` and `*`. After that the different pages of the domain are tested leading to an `AssertionError` for the following lines:
```
assertTrue(rules.isAllowed("http://www.domain.com/d"));
assertTrue(rules.isAllowed("http://www.domain.com/d/a"));
assertTrue(rules.isAllowed("http://www.domain.com/e/d/foo.html"));
assertTrue(rules.isAllowed("http://www.domain.com/foo/bar/baz.html"));
```
But from my point of view those pages are prohibited for the given agent names. The first three pages because of `Agent2` and `Agent3` and the last one because of `*`.

What do you think?

### Disadvantage

The changes I made to the parse class have a small disadvantage. (If I understood it correctly) the old implementation searched for a user-agent line, that matched the name of the agent. If such a user-agent has been found and all rule lines for this agent have been parsed, the parser was exiting the parsing routine if it found an additional user-agent with a different name than the agent name. This was simply caused by the idea, that for every user agent only one set of rules would be defined. Since I experienced the opposite, I had to delete this mechanism, forcing the parser to always parse the whole file.

Cheers,
Michael

### DIFF
--- a/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
+++ b/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
@@ -18,7 +18,6 @@
 package crawlercommons.robots;
 
 import java.io.UnsupportedEncodingException;
-import java.net.URI;
 import java.net.URL;
 import java.net.URLDecoder;
 import java.util.HashMap;
--- a/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
+++ b/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
@@ -87,9 +87,24 @@ public class SimpleRobotRulesParser extends BaseRobotsParser {
     }
 
     private static class ParseState {
+        /**
+         * Flag indicating whether the given name of the agent matched.
+         */
         private boolean _matchedRealName;
+        /**
+         * Flag that is true if not the given agent name but a wildcard matched.
+         */
         private boolean _matchedWildcard;
+        /**
+         * Flag that is true as long as it is allowed to add rules for the given
+         * agent.
+         */
         private boolean _addingRules;
+        /**
+         * Flag indicating whether all consecutive agent fields has been seen.
+         * It is set to false if an agent field is found and back to true if
+         * something else has been found.
+         */
         private boolean _finishedAgentFields;
 
         private String _url;
@@ -310,7 +325,8 @@ public class SimpleRobotRulesParser extends BaseRobotsParser {
         String encoding = "us-ascii";
 
         // Check for a UTF-8 BOM at the beginning (EF BB BF)
-        if ((bytesLen >= 3) && (content[0] == (byte) 0xEF) && (content[1] == (byte) 0xBB) && (content[2] == (byte) 0xBF)) {
+        if ((bytesLen >= 3) && (content[0] == (byte) 0xEF) && (content[1] == (byte) 0xBB)
+                && (content[2] == (byte) 0xBF)) {
             offset = 3;
             bytesLen -= 3;
             encoding = "UTF-8";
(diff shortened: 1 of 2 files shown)


lewismc:
I am struggling slightly to interpret whether the root issue (user-agent being defined more than once) is 'illegal' or just undesired. You've done well to describe your own issue but I am still struggling to see which method is better and/or more suitable. Can anyone else comment here? @MichaelRoeder thank you for PR.

MichaelRoeder:
I took a deeper view into the problem. The website http://www.robotstxt.org/ gives two references for the robots.txt file:
* *the original 1994 [A Standard for Robot Exclusion document](http://www.robotstxt.org/orig.html)*
* *a 1997 Internet Draft specification [A Method for Web Robots Control](http://www.robotstxt.org/norobots-rfc.txt)*

Both documents do not give information about this problem. But in the [Appendix of HTML4](http://www.w3.org/TR/html4/appendix/notes.html#h-B.4.1.1) I found this sentence:
*If the value is "***", the record describes the default access policy for any robot that has not matched any of the other records. It is not allowed to have multiple such records in the "/robots.txt" file.*

Thus, it seems that the example of http://datahub.io/robots.txt is invalid. From this point of view, this pull request is not needed anymore and should be closed.
Sorry, that I didn't found this sentence earlier. It might have saved us some time...

However, I took a closer look at the JUnit test Problem that I described in this pull request description. For your implementation of the parser the test is green, because the parser gets an array of names, that he should search for, and only uses the first hit found in the robots.txt. Thus, in this example the parser returns the rules for the agent `Agent1` and ignores all other given names. This means, that if a parser gets more than one agent name, the rules that are derived depend on the order of agent records inside the robots.txt file.
This **might** be the desired behaviour, but I think it should be made clear, e.g., by adding a javadoc description that describes this behaviour.

Cheers,
Michael

kkrugler:
Hi Michael - I just updated the Javadocs for the BaseRobotsParser class. Please take a look, and close this issue if the new docs clear things up.

MichaelRoeder:
The updated comment for the BaseRobotsParser class is very nicely written and makes the behaviour much clearer. Thanks for your work!

From my point of view only one small point is missing that might be added to the SimpleRobotRulesParser. If this parser discovers a rule set for a user agent with a matching name this set is parsed and returned as result. It doesn't matter how many other agent names the parser has got - it always takes the first matching set of rules. Thus, while the order of the agent names given to the parser does not matter, the order of the agent names inside the robots.txt file does.

Cheers,
Michael

lewismc:
Hi Michael,
Thank you so much for the feedback.
Are you able to provide another patch for this update? I think it would be
valuable.
Thanks
Lewis

On Wed, Sep 9, 2015 at 2:12 AM, Michael Röder <notifications@github.com>
wrote:

> The updated comment for the BaseRobotsParser class is very nicely written
> and makes the behaviour much clearer. Thanks for your work!
>
> From my point of view only one small point is missing that might be added
> to the SimpleRobotRulesParser. If this parser discovers a rule set for a
> user agent with a matching name this set is parsed and returned as result.
> It doesn't matter how many other agent names the parser has got - it always
> takes the first matching set of rules. Thus, while the order of the agent
> names given to the parser does not matter, the order of the agent names
> inside the robots.txt file does.
>
> Cheers,
> Michael
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/crawler-commons/crawler-commons/pull/91#issuecomment-138848811>
> .
>



-- 
*Lewis*

MichaelRoeder:
Hi all,

I added an explaining javadoc comment to the class. I thought it would be best to create a new pull request for this (#95) because otherwise I would have to revert my other commits.

Thanks for the discussion. I hope that it didn't created to much effort on your side. At least for me it was very interesting :)

Cheers,
Michael

PR #94, opened 2015-09-06
Upgrade to Jdk1.7

lewismc:
Hi Folks,
This PR updates pom.xml to specify javac 1.7 as well as Maven compilation for 1.7.
It then implements all of the localization and charset specification required to comply with the forbidden API's plugin. Additionally it removes all references to System.out.blah which have been replaced by String exception messages and typically wrapped in slf4j logging instances.
Thanks

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,22 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
-  Licensed to the Apache Software Foundation (ASF) under one
-  or more contributor license agreements.  See the NOTICE file
-  distributed with this work for additional information
-  regarding copyright ownership.  The ASF licenses this file
-  to you under the Apache License, Version 2.0 (the
-  "License"); you may not use this file except in compliance
-  with the License.  You may obtain a copy of the License at
-  
-    http://www.apache.org/licenses/LICENSE-2.0
-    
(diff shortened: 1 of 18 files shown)


lewismc:
Bugger i just noticed that this PR also includes the bloody commits for the validSitemaps branch as well. @Chaiavi can you please merge #82 into master and I will rebase? Thanks.

lewismc:
This PR is now good folks, it's been updated to master.

Chaiavi:
Thanks Lewis

I ran the unit tests and all passed

lewismc:
OK, will wait maybe another day and merge if all good.

lewismc:
Hi @jnioche PR updated to accommodate your suggestions. Quick note on existing String.format usage... we only use it within function calls which return string values. We cannot implementing slf4j logging conventions in such instances. 
I've implemented slf4j conventions in a few other places as well.

jnioche:
> we only use it within function calls which return string values. We cannot implementing slf4j logging conventions in such instances. 

sure, that's not what I was suggesting. Thanks for your work on this

kkrugler:
Hi @lewismc - I didn't see this mentioned in CHANGES.txt

lewismc:
Correct Ken. I used the merge function in the UI and forgot about this. Will update with adhoc commit right now.

Issue #92, opened 2015-09-01
Deprecate HTTP fetcher support


kkrugler:
In researching a bug in Bixo, I realized that the SimpleHttpFetcher needs to be serializable so that we can easily use it with Hadoop jobs. But that's an odd dependency, and in researching how we use the fetching code in Bixo, it felt like SimpleHttpFetcher (which effectively wraps Apache's HttpClient code) is too specific to put into crawler-commons. Plus there's a lot of deferred maintenance work in that code, to keep it in sync with HttpClient (e.g. many methods it uses are now deprecated).

So I feel like we should deprecate this, and the RobotUtils createFetcher() and getRobotRules() methods.

Let me know if you feel strongly that we should continue supporting the fetcher code, otherwise I'll deprecate it soon, and file an issue to remove it after a few releases.

jnioche:
+1 to your suggestion Ken

Chaiavi:
My intuition is that crawler-commons should continue supporting it.

@Ken - why do you suggest to deprecate it? - is it because we can expect
the user to use Apache's HttpClient ?

My intuition is that if we create a common library for crawling, the main
component in it should be the actual page fetcher, if we ditch that then
what do we leave in our project ? - only the robots handling and the
Sitemap parser ?


Although I agree that it is a high maintenance component as Igor seems to
not prioritize backwards compatibility...

kkrugler:
We want to have code that's common across _all_ crawlers. There are several HTTP libraries that crawlers could use, of which HttpClient is just one. So for that reason alone I wasn't confident it made sense to add it in the first place. And then how you configure the fetcher depends a lot on your use case - e.g. do you need to handle cookie caching? And finally there's the issue of keeping it in sync with HttpClient.

As for what should be in crawler-commons, here's the page from the original meeting (Nov 4th, 2009 at ApacheCon) where we first discussed this project: http://wiki.apache.org/nutch/ApacheConUs2009MeetUp

What's not yet in crawler-commons that I think would be useful is:

 - URL normalization
 - URL filtering
 - Page similarity (for detecting real page changes, and/or page deduplication)
 - Testing code (e.g. extract bixo code that synthesizes web pages from a real-world link graph)

I think think of other things like a DB of link shortening domains, similar to what we do for TLD support.

Chaiavi:
Thank you for the link and explanations

Now I am still not for deprecating it but not against either, I trust your
judgement.

jnioche:
[...]

PR #97, opened 2015-09-14
Deprecate fetcher classes

kkrugler:
This fixes issue #92


jnioche:
[master f3f3484] Deprecate fetcher classes #97

PR #100, opened 2015-10-05
updated tld names from publicsuffix.org

jnioche:


lewismc:
+1 Julien

jnioche:
Thanks @lewismc

Issue #101, opened 2015-10-06
Implement log4j.properties for main resources


lewismc:
it is always nice to be able to log code execution. This issue simply provides a log4j.properties exactly for that.

PR #102, opened 2015-10-06
Implement log4j.properties for main resources #101 

lewismc:
This PR addresses #101

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,7 @@
 Crawler-Commons Change Log
 
 Current Development 0.7-SNAPSHOT (yyyy-mm-dd)
+- Issue 101: Implement log4j.properties for main resources (lewismc)
 - Issue 89: Upgraded Tika 1.10 (jnioche)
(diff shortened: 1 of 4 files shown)


kkrugler:
Hi Lewis - I don't think we want to do this, as the crawler-commons jar is used as a dependency by projects which want to make their own decision on what logging framework (and log4j.properties) to use. By changing the scope of the slf4j-log4j12 jar from test to scope, you've now created a transitive dependency that will pull that jar in for any project using crawler-commons.

If the issue is that the SiteMapTester tool doesn't have logging, then this should be packaged as a separate uber-jar, as per my comment (June 30, 2014) on the list, where I'd said:

>It would be cleanest if this was a separate artifact from the build - e.g. we create a crawler-commons jar, and a crawler-commons-tools.jar, where the latter is an uber jar (includes all dependencies) so you can just run it from the command line.

So I'm -1 on this.

lewismc:
Ah OK Ken. Makes good sense. I'll package this into a separate tool and submit a PR.

Issue #74, opened 2015-04-22
Provide class for basic URL normalization 


jnioche:
Similar to the BasicURLNormalizer in Nutch with maybe more later on. 

Related to #24

Chaiavi:
[URL Normalization specs from Wikipedia] (https://en.wikipedia.org/wiki/URL_normalization)

Candidates for initial class:
[Nutch Basic Normalizer] (https://github.com/apache/nutch/blob/trunk/src/plugin/urlnormalizer-basic/src/java/org/apache/nutch/net/urlnormalizer/basic/BasicURLNormalizer.java)

[Crawler4j Normalizer] (https://github.com/yasserg/crawler4j/blob/master/src/main/java/edu/uci/ics/crawler4j/url/URLCanonicalizer.java)

jnioche:
See related thread on [https://github.com/DigitalPebble/storm-crawler/issues/120]

Chaiavi:
Adding two more candidates to the mix:
[Bixo](https://github.com/bixo/bixo/blob/master/src/main/java/bixo/urls/SimpleUrlNormalizer.java)
[StormCrawler](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/storm/crawler/filtering/basic/BasicURLNormalizer.java)

Chaiavi:
It seems there are "must have" features of URL Normalization.

And there are optional features which not everyone thinks they should be in a normalizer.

I suggest having the URLNormalizer do the "Must have"s, and have additional methods (using method chaining?) for the optional features (like removing directory index, removing "www" - which might change the actual page and not everyone likes)

jnioche:
Changed the title now that we do have some basic resources. This issue can be used to track possible improvements.

PR #106, opened 2015-11-10
Added URLFilter interface + BasicURLNormalizer borrowed from Nutch

jnioche:
Related to #74

This PR adds a very simple URLFilter interface with a single method `filter(String URL)` which returns null  if the URL should be filtered out or a modified version of it if it is normalised. This is similar to the approach I used in [storm-crawler](https://github.com/DigitalPebble/storm-crawler/wiki/URLFilters) and is something we meant to do in Nutch a long time ago.

This also contains a modified version of Nutch's BasicURLNormalizer. Later on we could add regex-based URLFilters and normalizers.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,7 @@
 Crawler-Commons Change Log
 
 Current Development 0.7-SNAPSHOT (yyyy-mm-dd)
+- Issue 106: Added URLFilter interface + BasicURLNormalizer (jnioche)
 - Issue 89: Upgraded Tika 1.10 (jnioche)
 - Issue 82: [Sitemaps] Upgrade Valid / Legal / Strict SitemapUrls (Avi Hayun)
 - Issue 60: [Sitemaps] Upgrade Valid / Legal / Strict SitemapUrls (Avi Hayun)
--- a/src/main/java/crawlercommons/filters/URLFilter.java
+++ b/src/main/java/crawlercommons/filters/URLFilter.java
@@ -1,11 +1,28 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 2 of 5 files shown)


Chaiavi:
Thanks Julien,

Just skimmed and it looks like a good baseline (although I really didn't delve into the code).

I saw a "Main" method in the Basic Normalizer class, maybe we should have it as a "tool" like you did in the sitemaps package ?

jnioche:
Thanks for your comments @Chaiavi 

> I saw a "Main" method in the Basic Normalizer class, maybe we should have it as a "tool" like you did in the sitemaps package ?

yes, it is probably temporary. We could have a generic URLfilter tester class later once we have more than one flavour of filter and a mechanism for declaring them, like the one in [StormCrawler](https://github.com/DigitalPebble/storm-crawler/blob/master/core/src/main/java/com/digitalpebble/storm/crawler/filtering/URLFilters.java) for instance.

kkrugler:
Hi @jnioche,

Looks OK to roll in, though I prefer abstract classes over interfaces. That way if we need to extend it, we don't break every implementation. For example, if we wanted a new API that let the caller specify things like whether to treat a ref (http://domain.com#blah) as a query parameter, we could add this without breaking things.

Separate note - see [SimpleUrlNormalizerTest.java](https://github.com/bixo/bixo/blob/master/src/test/java/bixo/urls/SimpleUrlNormalizerTest.java) in [Bixo](https://github.com/bixo/bixo) for a different set of test normalizations. It would be interesting to see how many of these pass the Nutch-based code.

jnioche:
Thanks @kkrugler 

re-abstract class \: I completely agree. Will change that. I had actually followed the same logic in [stormcrawler's ParsFilter](https://github.com/DigitalPebble/storm-crawler/commit/8e5556e357c1c75861652712938662ab6d9c10ce), should also do the same with the URLFilters there.

There's definitely more to be done on the normalization side. We had listed some things in #74 and #120, this is really work in progress.

jnioche:
Merged in [master c1b3f4b]

PR #98, opened 2015-09-25
Upgrades httpclient to version 4.5.1 and fixes deprecated methods

aecio:
This pull request upgrades httpclient library to version 4.5.1 and removes all deprecated methods usages from SimpleHttpFetcher. This class is in process of being deprecated, but this can still be useful and it would be good to have this in the next version.

PR #115, opened 2016-05-04
Fixes bug introduced in pull request #98

aecio:
Pull request #98 updated httpclient to version 4.5.1. But a small bug was introduced in which the settings being set to the SimpleHttpFetcher were not being used in the final HttpClient object. This PR fixes that problem and also allows the user to configure a new timeout introduced in the new httpclient version.

SimpleHttpFetcher is deprecated, but at least the next version will not introduce this bug if this gets merged.


kkrugler:
Merged - thanks Aécio!

Issue #36, opened 2015-04-09
Support Image Sitemap's


GoogleCodeExporter:
```
Google's documentation on SiteMap support highlights some anomalies with 
regards to the above SiteMap type.
We should consider supporting such functionality.

https://support.google.com/webmasters/answer/178636?hl=en&ref_topic=4581190


```

Original issue reported on code.google.com by `lewis.mc...@gmail.com` on 11 Apr 2014 at 7:21

jnioche:
More generally this is about how to handle extensions of the basic sitemap protocol. Google seem to have developed a whole range of them for news etc...  A source of inspiration could be how ROME  [http://rometools.github.io/rome/] deals with extensions of the feeds, see ForeignMarkup in  [https://github.com/rometools/rome/blob/master/src/main/java/com/rometools/rome/feed/WireFeed.java#L234]

kkrugler:
Rome handles this by returning a list of DOM elements (actual JDOM2 elements), so that it's able to handle pretty much anything that could be found as "extras". If we did the same, we'd be coercing whatever additional data we found in a sitemap into a DOM structure...does that seem OK?

Issue #35, opened 2015-04-09
Support Video Sitemap's


GoogleCodeExporter:
```
Google's documentation on SiteMap support states anomalies in the above 
SiteMap's which we should consider supporting.
https://support.google.com/webmasters/answer/80471?hl=en&ref_topic=4581190


```

Original issue reported on code.google.com by `lewis.mc...@gmail.com` on 11 Apr 2014 at 7:20

Issue #29, opened 2015-04-09
More robust parsing of sitemap index files


GoogleCodeExporter:
```
http://activision.taleo.net/careersection/sitemap.jss

contains entries such as 

<sitemap>
http://activision.taleo.net/careersection/sitemap.jss?portalCode=2&lang=en
</sitemap>

i.e. the mandatory <loc> element is missing

Ideally the site should produce the correct content following the schema but we 
should make the parser a bit more robust and produce outlinks even if the loc 
element is missing



```

Original issue reported on code.google.com by `digitalpebble` on 6 Sep 2013 at 9:28

PR #116, opened 2016-07-01
Parse XML sitemaps via SAX for option to parse partial/broken documents

jnioche:
Updated #80 - merged with master branch, fixed a few issues, applied formatting, fix headers etc...

Will compare the perfs on some large sitemap files. I'd expect if not a gain in speed, then at least one in memory usage (since we don't hold the whole DOM representation in RAM).

Have closed #80, let's resume the work here.

@matt-deboer sorry it took so long to have another look at your work


jnioche:
I ran a benchmark on this branch vs existing DOM parsing. As expected is uses a lot less memory (90MB vs 180MB on my test docs)

[https://www.novinky.cz/sitemap/sitemap_articles_5.xml]
[http://www.foxnews.com/sitemap.xml?idx=26]
[http://push.abs-cbn.com/sitemap/2015/]

src/main/java/crawlercommons/sitemaps/AbstractSiteMapSAXHandler.java
```diff
@@ -0,0 +1,130 @@
...
+        // the <rss> or <rdf> tag is a <channel> tag, so we can use that.
+        // See https://github.com/crawler-commons/crawler-commons/issues/87
+        // and also RSS 1.0 specification http://web.resource.org/rss/1.0/spec
+        else if ("channel".equals(qName)) {
+            delegate = new SiteMapRSSSAXHandler(url, elementStack, strict);
+        }
+    }
+
+    public void endElement(String uri, String localName, String qName) throws SAXException {
+        if (delegate != null)
```
kkrugler: add code block for consistency

src/main/java/crawlercommons/sitemaps/AbstractSiteMapSAXHandler.java
```diff
@@ -0,0 +1,130 @@
...
+import org.xml.sax.Attributes;
+import org.xml.sax.SAXException;
+import org.xml.sax.SAXParseException;
+import org.xml.sax.helpers.DefaultHandler;
+
+/**
+ * Provides a base SAX handler for parsing of XML documents representing
+ * sub-classes of AbstractSiteMap.
+ */
+public class AbstractSiteMapSAXHandler extends DefaultHandler {
```
kkrugler: Seems odd to call this class AbstractXXX when it's not abstract.

src/main/java/crawlercommons/sitemaps/AbstractSiteMapSAXHandler.java
```diff
@@ -0,0 +1,130 @@
...
+            delegate.characters(ch, start, length);
+        }
+    }
+
+    protected String currentElement() {
+        return elementStack.peek();
+    }
+
+    protected String currentElementParent() {
+        return (elementStack.size() < 2) ? null : elementStack.get(1);
```
kkrugler: Shouldn't this be a get(<list size> - 2), not get(1)?

src/main/java/crawlercommons/sitemaps/SiteMapIndexSAXHandler.java
```diff
@@ -0,0 +1,104 @@
...
+        if (loc != null) {
+            SiteMap s = new SiteMap(loc, lastMod);
+            sitemap.addSitemap(s);
+            LOG.debug("  {}. {}", (i + 1), s);
+        }
+        loc = null;
+        lastMod = null;
+    }
+
+    public void error(SAXParseException e) throws SAXException {
```
kkrugler: Why do we maybe want to add the sitemap if we get an error while parsing? Is this part of the "handle broken sitemaps" logic? If so, a comment would be good.

src/main/java/crawlercommons/sitemaps/SiteMapRSSSAXHandler.java
```diff
@@ -0,0 +1,153 @@
...
+
+    /*
+     * (non-Javadoc)
+     * 
+     * @see
+     * crawlercommons.sitemaps.AbstractSiteMapSAXHandler#startElement(java.lang
+     * .String, java.lang.String, java.lang.String, org.xml.sax.Attributes)
+     */
+    @Override
+    public void startElement(String uri, String localName, String qName, Attributes attributes) throws SAXException {
```
kkrugler: Doesn't this need to call super.startElement? Otherwise the element stack isn't going to be set up, right?

src/main/java/crawlercommons/sitemaps/SiteMapSAXHandler.java
```diff
@@ -0,0 +1,118 @@
...
+    boolean valid;
+    private int i = 0;
+
+    SiteMapSAXHandler(URL url, LinkedList<String> elementStack, boolean strict) {
+        super(elementStack, strict);
+        sitemap = new SiteMap(url);
+        sitemap.setType(SitemapType.XML);
+    }
+
+    public void startElement(String uri, String localName, String qName, Attributes attributes) throws SAXException {
```
kkrugler: Doesn't super.startElement() need to be called? Or (confusingly) is this class only used as a delegate from AbstractSiteMapSAXHandler? Something seems odd here.

src/main/java/crawlercommons/sitemaps/AbstractSiteMapSAXHandler.java
jnioche: I agree. Would DefaultSiteMapSAXHandler be a better name? or rename SiteMapSAXHandler into  something else and use SiteMapSAXHandler for this one?

src/main/java/crawlercommons/sitemaps/AbstractSiteMapSAXHandler.java
jnioche: do you mean `get(getSize() - 2)` ? Makes sense.

src/main/java/crawlercommons/sitemaps/AbstractSiteMapSAXHandler.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    public AbstractSiteMap getSiteMap() {
```
jnioche: Overlooked I suppose

jnioche:
The SAX parsers do not work for cases like #29. Will add a test class to the master branch and apply it to this branch as well + fix some of the issues highlighted by @kkrugler.

src/main/java/crawlercommons/sitemaps/SiteMapIndexSAXHandler.java
jnioche: I can only second guess what @matt-deboer intended. It could be to make sure the loc and lastmod fields are reset to null, however a new instance of the handler class would be created for a  new doc anyway. It could help with the "handle broken sitemaps" if the exception occurred between the discovery of the location and the next sitemap or sitemapindex tag.

src/main/java/crawlercommons/sitemaps/SiteMapSAXHandler.java
jnioche: I think this is these classes are used only as delegates.

src/main/java/crawlercommons/sitemaps/AbstractSiteMapSAXHandler.java
jnioche: actually the existing code is correct. The more recent elements are added to the head of the queue so the element parent will always be in position 1. For some reason I expected them to be put at the end.

src/main/java/crawlercommons/sitemaps/AbstractSiteMapSAXHandler.java
jnioche: Fixed in b9d5128

src/main/java/crawlercommons/sitemaps/AbstractSiteMapSAXHandler.java
jnioche: b9d5128 -> renamed the classes. @kkrugler let me know what you think of it

src/main/java/crawlercommons/sitemaps/AbstractSiteMapSAXHandler.java
jnioche: Fixed in b9d5128

jnioche:
Have identified a possible issue with the locations not being properly handled. Will add a test case to the master branch first.

@kkrugler I have addressed some of the issues you mentioned, will look again at the startElement() logic.

jnioche:
Was an issue with locations split over several calls to characters() - as illustrated by d200346

Still need to apply the same logic to XMLIndex, Atom and RSS Handlers and modify tests to cover for these cases.

The latest commit also improves the efficiency of the parsing as URLs were created twice from strings unnecessarily.

sebastian-nagel:
I've adapted this PR to master and for testing and evaluation doubled the parser classes (added SiteMapParserSAX), see [this fork](https://github.com/sebastian-nagel/crawler-commons/tree/cc-116-sitemap-sax-parser). After testing both classes on 35 problematic sitemaps and feeds: I've found no regressions, truncated docs are successfully parsed as expected, XML syntax errors do not affect the whole document, but only the tail after the error.

There are two option how to proceed:
1. replace the existing DOM-based parser
2. keep both parsers and let the user decide which one to use (cf. [--use-sax-parser in SiteMapTester](https://github.com/sebastian-nagel/crawler-commons/blob/cc-116-sitemap-sax-parser/src/main/java/crawlercommons/sitemaps/SiteMapTester.java))

kkrugler:
Hi @sebastian-nagel - so what would be a reason for keeping the DOM-based parser around? From @jnioche's comments above, SAX should be faster, use less memory, and should be more resilient to errors (would this also take care of #148?).

sebastian-nagel:
For the long-term, I principally agree: maintaining two parsers means additional work, and there are few arguments in favor of the DOM-based parser:
- less code, 700 vs. 1000 lines, excluding license headers, ev. easier to read
- optionally expose the DOM or parts of it? For the more complex, derived formats (images #36, videos #35, news) this could be a quick and extensible solution to make the additional information accessible without the need to define huge classes to hold all the information.

> should be more resilient to errors

Right now, the only difference is that the SAX parser is able to extract URLs from the top of the XML document. It does not recover from an error. Would be nice if the parser could continue after trivial errors such as in #148. I don't know whether this is doable based on the existing code. I found one example ([1](https://gist.github.com/chrisguitarguy/1305010), [2](http://blog.viktorpetersson.com/post/115153423414/how-to-parse-and-dump-a-sitemap)) in Python based on [beautifulsoup](https://pypi.python.org/pypi/beautifulsoup4).

> SAX should be faster, use less memory

In addition, it could provide an iterator over SiteMapURLs. Often URLs are filtered or sampled, and there is no need to keep the complete list of 50,000 URLs (or more for sitemap indexes) in memory. 

For the next time, I want to test the SAX-based parser and compare it to the DOM-based one. I'll prepare a larger test set (few 1000 sitemaps), right now I can share a list of 16 million URLs pointing to sitemaps.

jnioche:
+1 to having both for the time being.  @sebastian-nagel could you please open a new PR?

sebastian-nagel:
Done: #150. I've used the SAX parser to process 500,000 sitemaps, but have to mine the logs for potential regressions.

jnioche:
>  I've used the SAX parser to process 500,000 sitemaps

did you monitor the processing time and memory used compared to the DOM based one?

sebastian-nagel:
No, it was a mapreduce job, and also only an overlapping set of URLs. For performance measures, it's probably also better to use a sample on disk not URLs. A first comparison of the log files also shows that the Tika version (also changed) has some impact whether formats are properly recognized. I'll sample a subset of URLs (excluding sitemap indexes) and pre-fetch it for measuring performance, and to have a controlled environment with no differences except the parser.

jnioche:
Replaced by #150

PR #120, opened 2016-09-12
Faster parsing of dates. Fixes #117

jnioche:
Fixes the format for ISO dates and tries using it before DatatypeConverter. This improves the parsing time by 30 to 40%.

Before (average over 10 iterations in msec)

Average for http://www.foxnews.com/sitemap.xml?idx=26 => 684
Average for http://push.abs-cbn.com/sitemap/2015/ => 238
Average for https://www.novinky.cz/sitemap/sitemap_articles_5.xml => 972

After

Average for http://www.foxnews.com/sitemap.xml?idx=26 => 481
Average for http://push.abs-cbn.com/sitemap/2015/ => 141
Average for https://www.novinky.cz/sitemap/sitemap_articles_5.xml => 657

I also tried java.time (in Java 8) but there was no benefit from it
           ` return Date.from(java.time.ZonedDateTime.parse(date).toInstant());`

Average for http://www.foxnews.com/sitemap.xml?idx=26 => 646
Average for http://push.abs-cbn.com/sitemap/2015/ => 197
Average for https://www.novinky.cz/sitemap/sitemap_articles_5.xml => 1281

PR #121, opened 2016-09-13
Adding support for Google News Sitemaps

michaellavelle:


jnioche:
thanks @michaellavelle 

We discussed in #36 whether we could have a generic way of handling various extensions instead of hardcoding a mechanism for every single one of them. Any thoughts on that? How would the news extension fit into that?

michaellavelle:
@jnioche Thanks for your message, and for this really useful project!.  A generic way of handling extensions would definitely be a good way to go - some form of SiteMapFactory?   In the meantime,  is this PR something you'd be considering for merge.   This functionality is something we require in the near future and due to the existing private methods of SiteMapParser, it's not possible for us to create our own custom subclass.    An alternative could be to make the methods of SiteMapParser protected - allowing users of the project to add this functionality to their own subclasses - wondered what your thoughts were on this?
Also, would you have any plans for a release soon? - either one including this PR, or for a change to the protected methods - until then we don't have a way to parse Google News Sitemaps using crawler-commons.

jnioche:
Thanks @michaellavelle.

> An alternative could be to make the methods of SiteMapParser protected - allowing users of the project to add this functionality to their own subclasses - wondered what your thoughts were on this?

I would definitely +1 this. Would be a good way of doing until we have a more generic way of dealing with extensions.

Any chance you could contribute a PR for it?

jnioche:
@michaellavelle re-release, things have stalled a bit lately but it would be good to have a release soon. Will probably include #120 (which you are welcome to review if you can) as well as the change to 'protected' we mentioned.

jnioche:
@michaellavelle closing this as I have merged #124

Issue #125, opened 2016-09-25
Upgrade to JDK 1.8


lewismc:
I just tried to build CC with JDK 1.8 
```
lmcgibbn@LMC-056430 /usr/local/crawler-commons(master) $ java -version
java version "1.8.0_91"
Java(TM) SE Runtime Environment (build 1.8.0_91-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.91-b14, mixed mode)
```
This results in the following errors when generating Javadoc
```
lmcgibbn@LMC-056430 /usr/local/crawler-commons(master) $ mvn javadoc:javadoc
[INFO] Scanning for projects...
[WARNING] The POM for org.eclipse.m2e:lifecycle-mapping:jar:1.0.0 is missing, no dependency information available
[WARNING] Failed to retrieve plugin descriptor for org.eclipse.m2e:lifecycle-mapping:1.0.0: Plugin org.eclipse.m2e:lifecycle-mapping:1.0.0 or one of its dependencies could not be resolved: Failure to find org.eclipse.m2e:lifecycle-mapping:jar:1.0.0 in https://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] Building Crawler-commons 0.7-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[WARNING] The POM for org.eclipse.m2e:lifecycle-mapping:jar:1.0.0 is missing, no dependency information available
[WARNING] Failed to retrieve plugin descriptor for org.eclipse.m2e:lifecycle-mapping:1.0.0: Plugin org.eclipse.m2e:lifecycle-mapping:1.0.0 or one of its dependencies could not be resolved: Failure to find org.eclipse.m2e:lifecycle-mapping:jar:1.0.0 in https://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced
[INFO]
[INFO] >>> maven-javadoc-plugin:2.10.4:javadoc (default-cli) > generate-sources @ crawler-commons >>>
[INFO]
[INFO] --- maven-enforcer-plugin:1.0:enforce (enforce-maven) @ crawler-commons ---
[INFO]
[INFO] <<< maven-javadoc-plugin:2.10.4:javadoc (default-cli) < generate-sources @ crawler-commons <<<
[INFO]
[...]

PR #126, opened 2016-09-26
Upgrade to JDK 1.8

lewismc:
This issue addresses #125

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: java
 
+jdk:
+  - oraclejdk8
+
 script:
-  - jdk_switcher use oraclejdk8
   - mvn install javadoc:aggregate
 
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: java
 
+script:
+  - jdk_switcher use oraclejdk8
+  - mvn install javadoc:aggregate
+
 notifications:
   email:
     - crawler-commons@googlegroups.com
--- a/src/main/java/crawlercommons/domains/EffectiveTldFinder.java
+++ b/src/main/java/crawlercommons/domains/EffectiveTldFinder.java
@@ -108,9 +108,11 @@ public class EffectiveTldFinder {
     }
 
     /**
-     * @param hostname the hostname for which to find the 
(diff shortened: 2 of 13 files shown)


lewismc:
For some reason the TravisCI config is not being picked up.

.travis.yml
```diff
@@ -1,5 +1,9 @@
 language: java
 
+script:
+  - jdk_switcher use oraclejdk8
```
jnioche: Have you tried 

```
jdk:
  - oraclejdk8
```

.travis.yml
```diff
@@ -1,5 +1,9 @@
...
+  - mvn install javadoc:aggregate
```
jnioche: what's this one for?

src/main/java/crawlercommons/domains/EffectiveTldFinder.java
```diff
@@ -74,13 +74,13 @@ public static EffectiveTldFinder getInstance() {
         return instance;
     }
 
-    public boolean initialize(InputStream effective_tld_data_stream) {
```
jnioche: Are _ not allowed in Java8?

jnioche:
Did you run `mvn java-formatter:format` before committing?

.travis.yml
lewismc: This defines the targets to run... we always wish to build the Javadoc target as this lets us know if the Javadoc is JDK8 compliant.

src/main/java/crawlercommons/domains/EffectiveTldFinder.java
lewismc: Yes... but they are flagged as warn's for the Javac. It is not Java convention.

lewismc:
> Did you run mvn java-formatter:format before committing?

Actually no I did not.. I will do that right now.
Thanks for review @jnioche

lewismc:
@jnioche can you please review again at your earlier opportunity. Thank you.

.travis.yml
jnioche: ok, thanks. I assumed the javadoc was checked by Travis

jnioche:
Merged, thanks @lewismc

PR #140, opened 2017-02-02
Increase size limit of sitemaps (10MB -> 50MB), fixes #132

### LINKED ISSUE #132: [Sitemaps] Update to the Sitemap official guidelines (50mb sitemap files)
An update has been issued allowing sitemaps to consist of 50mb instead of 10mb file size:
http://blogs.bing.com/webmaster/November-2016/Increasing-the-size-limit-of-Sitemaps-file-to-addr

The fix should be put into the SiteMapParser:
https://github.com/crawler-commons/crawler-commons/blob/master/src/main/java/crawlercommons/sitemaps/SiteMapParser.java

Where the line
public static final int MAX_BYTES_ALLOWED = 10485760;

Should be replaced with
public static final int MAX_BYTES_ALLOWED = 52428800;

I think that is the only fix, but I didn't check it thoroughly.

sebastian-nagel:


kkrugler:
Thanks!

PR #142, opened 2017-02-02
Use constants from StandardCharsets where applicable, fixes #141

### LINKED ISSUE #141: Use StandardCharsets where applicable
Java 7 introduced constants for universally available encodings ([StandardCharsets](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html)). These should be used instead of string constants ("UTF-8", "us-ascii", etc.) to avoid unnecessary UnsupportedEncodingExceptions. See also #63.

sebastian-nagel:
- replace String constants defining charsets by StandardCharset constants
  if available for given encoding and supported by called method
- remove unnecessary exceptions (UnsupportedEncodingException)

### DIFF
--- a/src/test/java/crawlercommons/sitemaps/SiteMapParserTest.java
+++ b/src/test/java/crawlercommons/sitemaps/SiteMapParserTest.java
@@ -16,13 +16,14 @@
 
 package crawlercommons.sitemaps;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
 import java.net.URL;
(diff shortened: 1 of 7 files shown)


src/test/java/crawlercommons/sitemaps/SiteMapParserTest.java
```diff
@@ -16,13 +16,14 @@
...
 package crawlercommons.sitemaps;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
```
kkrugler: Why is this new import needed?

sebastian-nagel: It's not need, correct.

PR #150, opened 2017-02-23
Optional SAX Parser for Sitemaps

sebastian-nagel:
Provide a SAX-based parser for sitemaps, fixes #116
- use ```SiteMapParser sitemapParser = new SiteMapParserSAX(strict, allowPartialParse);```

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/SiteMapTester.java
+++ b/src/main/java/crawlercommons/sitemaps/SiteMapTester.java
@@ -32,10 +32,18 @@ public class SiteMapTester {
 
     private static final Logger LOG = LoggerFactory.getLogger(SiteMapTester.class);
     private static SiteMapParser parser = new SiteMapParser(false);
+    private static SiteMapParser saxParser = new SiteMapParserSAX(false, true);
 
     public static void main(String[] args) throws IOException, UnknownFormatException {
         if (args.length < 1) {
+            LOG.error("Fetch and process a sitemap (recursively if a sitemap index)");
             LOG.error("Usage: SiteMapTester <URL_TO_TEST> [MIME_TYPE]");
+            LOG.error("Options:");
+            LOG.error("  URL_TO_TEST  URL of sitemap");
+            LOG.error("  MIME_TYPE    force processing sitemap as MIME type,");
+            LOG.error("               bypass automatic MIME type detection");
+            LOG.error("Java properties:");
+            LOG.error("  sitemap.useSax  if true use SAX parser to process sitemaps");
         } else {
             URL url = new URL(args[0]);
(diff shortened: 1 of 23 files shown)


lewismc:
This looks like a really nice piece of code @sebastian-nagel 
Apart from some inconsistencies in the formatting of Javadoc, all I can see if that some new tests you've introduced are annotated as @ Ignore. Was that still intentional.
Nice work.

sebastian-nagel:
Thanks, the Javadoc is now formatted.
Yes, the ignored tests are intentional: I've kept both parsers and the tests as similar as possible to compare them side by the side. That was also on the premise that the SAX-based parser will replace the DOM-based parser: move the class and test files, fix the class name, done. If we want to keep the DOM-based parser forever, we probably should avoid the duplicated code and also make the tests generic, so that we can run it with both parsers.

lewismc:
ack

src/main/java/crawlercommons/sitemaps/SiteMapTester.java
```diff
@@ -32,37 +32,50 @@
...
+            parse(url, mt, useSaxParser);
         }
     }
 
     /**
      * Parses a Sitemap recursively meaning that if the sitemap is a
      * sitemapIndex then it parses all of the internal sitemaps
      */
-    private static void parse(URL url, String mt) throws IOException, UnknownFormatException {
+    private static void parse(URL url, String mt, boolean useSaxParser) throws IOException, UnknownFormatException {
```
sebastian-nagel: Ok, you're right, removing arguments can break things. I'll update the PR.
But I wouldn't call SiteMapTester an API (the "tester" indicates it): it's not thread-safe due to the usage of a static parser object. If we want to keep a stable API after removing the DOM-based parser and rename SiteMapParserSAX to SiteMapParser, a factory which is configured by properties, could be the better way.

jnioche: > I wouldn't call SiteMapTester an API

Neither would I. I thought by looking through the commits that there was another change on the API proper. Not the case, as you pointed out.

sebastian-nagel: Updated PR, usage of the SAX parser is now triggered by a Java property.

jnioche:
thanks @sebastian-nagel. Could you please squash the commits? I think we're good to go after that.

sebastian-nagel:
Commits until and including the merge commit (merge [matt-deboer-master](/crawler-commons/crawler-commons/tree/matt-deboer-master)) are squashed.

sebastian-nagel:
Thanks!

PR #157, opened 2017-03-19
Fix parsing of gzipped text sitemaps, fixes #143

### LINKED ISSUE #143: [Sitemaps] gzipped text files fail to parse
SiteMapParser fails with
```
Exception in thread "main" crawlercommons.sitemaps.UnknownFormatException: Error parsing XML for: http://fihh.de/urllist.txt.gz
```
if an gzip compressed text sitemap is parsed (e.g., http://fihh.de/urllist.txt.gz). The [sitemaps spec](https://www.sitemaps.org/protocol.html#otherformats) explicitely allows that "you may compress your Sitemap text file using gzip".

sebastian-nagel:
- detect gzip embedded media type to decide whether to parse as text or XML
- changed signature of processText(..., String) to processText(..., URL) to be in sync with other processXXX(...) methods
- verified solution on large collection of sitemaps (cf. [post in cc group](https://groups.google.com/forum/#!topic/crawler-commons/pOLsCVwRsxY)). Also errors when parsing gzipped sitemaps which are neither XML or text are now transparently handled during media type detection: a recursively gzipped XML sitemap and a gzipped HTML document.


src/main/java/crawlercommons/sitemaps/SiteMapParser.java
```diff
@@ -209,13 +209,30 @@ public AbstractSiteMap parseSiteMap(String contentType, byte[] content, URL url)
...
+                return processText(url, content);
             } else if (GZ_MEDIA_TYPES.contains(mediaType)) {
-                return processGzip(url, content);
-            } else {
-                mediaType = MEDIA_TYPE_REGISTRY.getSupertype(mediaType); // Check parent
-                return parseSiteMap(mediaType.toString(), content, url);
+                InputStream decompressed;
+                MediaType embeddedType;
+                try {
+                    decompressed = new GZIPInputStream(new ByteArrayInputStream(content));
```
jnioche: try-with-resources so that the stream gets closed no matter what happens?

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
```diff
@@ -292,22 +323,18 @@ protected SiteMap processText(String sitemapUrl, byte[] content) throws IOExcept
      */
     protected AbstractSiteMap processGzip(URL url, byte[] response) throws IOException, UnknownFormatException {
```
jnioche: rename the method to show that it is about xml e.g. `processedGzippedXML`

src/main/java/crawlercommons/sitemaps/SiteMapParserSAX.java
```diff
@@ -211,13 +211,30 @@ public AbstractSiteMap parseSiteMap(String contentType, byte[] content, URL url)
...
+                MediaType embeddedType;
+                try {
+                    decompressed = new GZIPInputStream(new ByteArrayInputStream(content));
+                    embeddedType = MediaType.parse(TIKA.detect(decompressed));
+                } catch (Exception e) {
+                    UnknownFormatException err = new UnknownFormatException("Failed to detect embedded MediaType of gzipped sitemap: " + url + ", caused by " + e);
+                    err.initCause(e);
+                    throw err;
+                }
+                if (XML_MEDIA_TYPES.contains(embeddedType)) {
```
jnioche: The repetition of code is not very elegant. What about having a separate recursive method for detecting the type instead and then an old fashioned switch statement to call the corresponding processX method? Any reason not to do that?

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
sebastian-nagel: +1

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
sebastian-nagel: A ByteArrayInputStream does not hold file handle which may leak, the stream objects are garbage collected anyway.

src/main/java/crawlercommons/sitemaps/SiteMapParserSAX.java
sebastian-nagel: I already tried such a solution and decided not to implement it after I've seen a [recursively compressed sitemap file](http://scores.sify.com/sitemap.xml.gz) which would require
- a counter to track and limit recursion to be save from zip bombs
- and also (or alternatively) a parser entry point which unwraps the sitemap recursively

I've found the unrolled solution more clear, esp. because recursion + while-loop looks somewhat transparent. Btw., I've removed the recursive call with the media super type - this is already handled by the while-loop.

jnioche: OK!

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
sebastian-nagel: Ok, PR updated.

jnioche:
Nice one! Thanks @sebastian-nagel

PR #165, opened 2017-06-06
SAX parser to stop URL at closing </loc> fixes #153

### LINKED ISSUE #153: [Sitemaps] SAX parser to stop URL at closing </loc>
The sitemaps SAX parser does not stop a URL at the closing `</loc>`. If there is no closing `</url>` the URL contains all concatenated text, including line breaks. 

Attached [two documents](https://github.com/crawler-commons/crawler-commons/files/824399/sitemaps-invalid-sax-parser.zip) to reproduce the problem. Since both do not validate against the [sitemap schema](http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd)), the behavior of the DOM parser is acceptable (fails resp. adds only a single URL).

Note: the solution should not break #29 (no `<loc>` element at all).

sebastian-nagel:
- on opening <loc> and <url> assume forgotten closing </url> tag
  only if there is more than white space in the buffer

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/sax/XMLHandler.java
+++ b/src/main/java/crawlercommons/sitemaps/sax/XMLHandler.java
@@ -72,7 +72,20 @@ class XMLHandler extends DelegatorHandler {
     public void startElement(String uri, String localName, String qName, Attributes attributes) throws SAXException {
         // flush any unclosed or missing URL element
         if (loc.length() > 0 && ("loc".equals(qName) || "url".equals(qName))) {
-            maybeAddSiteMapUrl();
+            // check whether loc isn't white space only
+            for (int i = 0; i < loc.length(); i++) {
+                if (!Character.isWhitespace(loc.charAt(i))) {
+                    maybeAddSiteMapUrl();
+                    return;
+                }
+            }
+            loc = new StringBuilder();
+            if ("url".equals(qName)) {
+                // reset also attributes
+                lastMod = null;
+                changeFreq = null;
+                priority = null;
+            }
         }
     }
 


src/main/java/crawlercommons/sitemaps/sax/XMLHandler.java
```diff
@@ -72,7 +72,20 @@
     public void startElement(String uri, String localName, String qName, Attributes attributes) throws SAXException {
         // flush any unclosed or missing URL element
         if (loc.length() > 0 && ("loc".equals(qName) || "url".equals(qName))) {
-            maybeAddSiteMapUrl();
+            // check whether loc isn't white space only
+            for (int i = 0; i < loc.length(); i++) {
```
jnioche: StringUtils.isNotBlank() would be more concise

jnioche: or the reverse i.e just return if StringUtils.isBlank(xxx)returns true

src/main/java/crawlercommons/sitemaps/sax/XMLHandler.java
```diff
@@ -72,7 +72,20 @@
...
+                if (!Character.isWhitespace(loc.charAt(i))) {
+                    maybeAddSiteMapUrl();
+                    return;
+                }
+            }
+            loc = new StringBuilder();
```
jnioche: we probably want to add it not just when a new URL element is found but also a loc (see line 74 above) so why not just call maybeAddSiteMapUrl?

src/main/java/crawlercommons/sitemaps/sax/XMLHandler.java
sebastian-nagel: Right, that's more concise. It adds commons-lang as new dependency.

src/main/java/crawlercommons/sitemaps/sax/XMLHandler.java
sebastian-nagel: A URL is added in both cases, as far as we have more than whitespace buffered in `loc`. But a call of maybeAddSiteMapUrl() turned out to be expensive. I didn't want to move the test for whitespace-only strings to this function to keep the difference between empty `<loc></loc>` and only a opening `<loc>` or `<url>`.

jnioche: I agree that the test for whitespace-only should be outside maybeAddSiteMapUrl(). It's the portion of code below which I had misread. No probs.

src/main/java/crawlercommons/sitemaps/sax/XMLHandler.java
jnioche: no big deal

sebastian-nagel: ok, I'll change the PR and commons-lang

jnioche: I thought we already had it but this is not the case apparently. Probably because we get it for free as a subdependency in so many projects that I expect it to be always here. No big deal, let's leave the code as it is.

jnioche:
thanks for the fix and the explanations @sebastian-nagel

src/main/java/crawlercommons/sitemaps/sax/XMLHandler.java
jnioche: too late. I've merged and I actually meant the opposite as you can see from my subsequent comment.

Issue #166, opened 2017-06-16
RSS parser ignores pubDate of link


MichealKum:
SiteMapParser.parseRSS uses pubDate of channel node as a lastMod for all items ignoring pubDate tag of item node.

lewismc:
hi @MichealKum ... can you propose a patch? A Unit test would be ideal ;)

MichealKum:
@lewismc sure

MichealKum:
@lewismc something like this:
[ISSUE-166_0.patch.zip](https://github.com/crawler-commons/crawler-commons/files/1081967/ISSUE-166_0.patch.zip)

lewismc:
Hi @MichealKum , sorry can you propose a pull request? Thank you

MichealKum:
@lewismc I unable to push my branch to the repository. Maybe I do it wrong?

MichealKum:
@lewismc sorry, now I know, how to do that. PR is created.

lewismc:
Excellent, PR is available at https://github.com/crawler-commons/crawler-commons/pull/167. Thank you, I'll review when I get a minute.

PR #167, opened 2017-06-20
ISSUE #166 RSS parser ignores item's pubDate

MichealKum:
* Converts RSS timestamp format to w3c format
* Use pubDate of item in new SiteMapURL creation

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/AbstractSiteMap.java
+++ b/src/main/java/crawlercommons/sitemaps/AbstractSiteMap.java
@@ -53,6 +53,38 @@ public abstract class AbstractSiteMap {
         }
     };
 
+    private static final ThreadLocal<DateFormat> W3C_FULLDATE_FORMAT_WITH_OFFSET = new ThreadLocal<DateFormat>() {
+        protected DateFormat initialValue() {
+            SimpleDateFormat result = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ", Locale.ROOT);
+            result.setTimeZone(TimeZone.getTimeZone("UTC"));
+            return result;
+        }
+    };
+
+    protected static final ThreadLocal<DateFormat> RFC822_FULLDATE_FORMAT = new ThreadLocal<DateFormat>() {
+        protected DateFormat initialValue() {
+            return new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss Z", Locale.ROOT);
+        }
+    };
+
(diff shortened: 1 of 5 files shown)


kkrugler:
Hi @MichealKum - thanks for the patch, looks good (nice unit tests, too). One question...is the set of date format parsers you created based on an RSS spec, or just what you've seen out in the wild? Wondering if we'll get a pull request in the future for additional date formats :)

MichealKum:
Hi @kkrugler , accordingly to the RFC 822 (RSS use this standart for pubDate field) the date-time field should have the format as following: `[ day "," ] date time`. It also says that a year should be a two-digit-year, but RSS files which I saw also use four digits. Moreover if you parse date with two-digit-year using `yyyy` then you will get 0017 for 17 instead of 2017. That is why there are these four formats.

src/main/java/crawlercommons/sitemaps/AbstractSiteMap.java
```diff
@@ -53,6 +53,38 @@ protected DateFormat initialValue() {
...
+        }
+    };
+
+    protected static final ThreadLocal<DateFormat> RFC822_SHORT_YEAR_DATE_FORMAT = new ThreadLocal<DateFormat>() {
+        protected DateFormat initialValue() {
+            return new SimpleDateFormat("EEE, dd MMM yy HH:mm:ss Z", Locale.ROOT);
+        }
+    };
+
+    protected static final ThreadLocal<DateFormat> RFC822_SHORTDATE_FORMAT = new ThreadLocal<DateFormat>() {
```
kkrugler: It would be good to include a comment (before all of these date formats) for the RFC 822 standard, and why we've got extra options. E.g. if the various options are with/without the day of week, and yy vs. yyyy, and with or without the UTC timezone, then it seems like we need 8 unique combinations - so some explanation for why there are only 5.

MichealKum: Actually only 4. The fifth one is for generating string representation for parsing by convertToDate. Not sure if it is necessary to cover 'with and without timezone'.
Also, probably it could be reasonable to use single ThreadLocal<DateFormat[]> instead of many new static fields.

kkrugler:
Thanks @MichealKum! I've merged. I'll also go add a line to CHANGES.txt (which I should have noted in my review, sorry) about your contribution.

MichealKum:
@kkrugler great. By the way: why don't you squash changes during merge? I think it is better to have one commit on master for each PR.

kkrugler:
Hi @MichealKum - yes, normally I squash...I forgot that you added a second commit.

PR #170, opened 2017-07-17
Sitemap SAX parser mangles sitemap URLs in sitemap index, fixes #169

### LINKED ISSUE #169: Sitemap SAX parsing mangles target URLs
http://stylecaster.com/sitemap.xml
gives 
http://stylecaster.com/sitemap.xml?yyyy=2017
where it should be 
http://stylecaster.com/sitemap.xml?yyyy=2017&mm=07&dd=11

the DOM parsers handles these correctly.

sebastian-nagel:
- completely add sitemap URLs from sitemap index if URL contains
  XML entities or CDATA
- update unit tests to cover this problem

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/sax/XMLIndexHandler.java
+++ b/src/main/java/crawlercommons/sitemaps/sax/XMLIndexHandler.java
@@ -54,7 +54,8 @@ import crawlercommons.sitemaps.AbstractSiteMap.SitemapType;
 class XMLIndexHandler extends DelegatorHandler {
 
     private SiteMapIndex sitemap;
-    private URL loc;
+    private StringBuilder loc;
+    private boolean locClosed;
     private Date lastMod;
     private int i = 0;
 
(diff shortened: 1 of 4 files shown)


src/main/java/crawlercommons/sitemaps/sax/XMLIndexHandler.java
```diff
@@ -72,30 +74,24 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
             maybeAddSiteMap();
         } else if ("sitemapindex".equals(currentElement())) {
             sitemap.setProcessed(true);
+        } else if ("loc".equals(currentElement())) {
+            locFound = true;
```
kkrugler: Would we want to set `locFound = false` in the `endElement()` method, for we get `</loc>`? Maybe it doesn't matter...

sebastian-nagel: No, it's mandatory to set it to true here. `locFound` means "a `<loc>...</loc>` element has been already found in the current `<sitemap>...</sitemap>`". The point is: there are sitemap indexes which just add the URLs without a surrounding`<loc>` (see [sitemap.index.xml](/crawler-commons/crawler-commons/blob/master/src/test/resources/sitemaps/sitemap.index.xml)). Maybe the name `locFound` isn't optimal?

sebastian-nagel:
Thanks! Renamed `locFound` to `locClosed` to avoid any confusions.

Issue #85, opened 2015-06-18
[Sitemaps] Add cross-submit feature into our project


Chaiavi:
In the official sitemaps documentation there is a reference about [Sitemaps & Cross Submits](http://www.sitemaps.org/protocol.html#location)

In a nutshell, it means that there is a way for a sitemap to reference urls from an other domain (as long as it gets validated in the other domain, and as long as it contains urls from one domain only per sitemap)

We should implement this functionality.

Chaiavi:
Good scenario:

Snippet from nytimes.com robots.txt: 

Sitemap: http://spiderbites.nytimes.com/sitemaps/www.nytimes.com/sitemap.xml.gz
Sitemap: http://www.nytimes.com/sitemaps/sitemap_news/sitemap.xml.gz
Sitemap: http://spiderbites.nytimes.com/sitemaps/sitemap_video/sitemap.xml.gz

kkrugler:
See https://github.com/crawler-commons/crawler-commons/issues/81 for a lengthy discussion of the above.

sebastian-nagel:
PR #515 implements a cross-submit validator which removes URLs failing the validation. I decided not to do a deep integration of the the cross-submit checks because it would require to pass the cross-submit host(s) to the XML handlers calling `SiteMapParser.urlIsValid(...)`. The implementation in a separate validator class is transparent and could be easily adapt to specific needs, e.g., if one wants to count the number of invalidated URLs.

PR #173, opened 2017-09-05
Add namespace aware DOM/SAX parsing for XML Sitemaps.

grokwich:
Add namespace aware DOM/SAX parsing for XML Sitemaps.  RSS and Atom parsing is also namespace aware, but finding elements has been left "relaxed" by only matching on the element "localName" (legacy behavior).

Here is test xml that was failing:
```XML
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<sit:urlset xmlns:image="http://www.google.com/schemas/sitemap-image/1.1" xmlns:video="http://www.google.com/schemas/sitemap-video/1.1" xmlns:sit="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.w3.org/1999/xhtml link.xsd http://www.google.com/schemas/sitemap-video/1.1 video.xsd http://www.sitemaps.org/schemas/sitemap/0.9 sitemap.xsd http://www.google.com/schemas/sitemap-image/1.1 image.xsd">
    <sit:url>
        <sit:loc>http://www.example.com/1</sit:loc>
        <sit:changefreq>daily</sit:changefreq>
    </sit:url>
    <sit:url>
        <sit:loc>http://www.example.com/2</sit:loc>
        <sit:changefreq>daily</sit:changefreq>
    </sit:url>
</sit:urlset>
```

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
+++ b/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
@@ -72,6 +72,7 @@ public class SiteMapParser {
      */
     public static final int MAX_BYTES_ALLOWED = 52428800;
 
+    
     /* Tika's MediaType components */
     private static final Tika TIKA = new Tika();
     private static final MediaTypeRegistry MEDIA_TYPE_REGISTRY = MediaTypeRegistry.getDefaultRegistry();
@@ -79,7 +80,8 @@ public class SiteMapParser {
     private static final List<MediaType> XML_MEDIA_TYPES = new ArrayList<>();
     private static final List<MediaType> TEXT_MEDIA_TYPES = new ArrayList<>();
     private static final List<MediaType> GZ_MEDIA_TYPES = new ArrayList<>();
-
+    
+    
     static {
         initMediaTypes();
     }
@@ -160,6 +162,7 @@ public class SiteMapParser {
         asmCopy.setLastModified(sitemap.getLastModified());
         return asmCopy;
     }
+
 
     /**
      * Parse a sitemap, given the content bytes and the URL.
@@ -352,16 +355,21 @@ public class SiteMapParser {
      *             {@link org.xml.sax.InputSource}
      */
     protected AbstractSiteMap processXml(URL sitemapUrl, InputSource is) throws UnknownFormatException {
-
         Document doc = null;
 
         try {
             DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+            
             // disable validation and avoid that remote DTDs, schemas, etc. are fetched
             dbf.setValidating(false);
+            
+            // support an explicitly named namespace. 
+            dbf.setNamespaceAware(true);
+
             dbf.setXIncludeAware(false);
             dbf.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
             DocumentBuilder db = dbf.newDocumentBuilder();
+            
             db.setEntityResolver(new EntityResolver() {
                 // noop entity resolver, does not fetch remote content
                 @Override
@@ -369,6 +377,7 @@ public class SiteMapParser {
                     return new InputSource(new StringReader(""));
                 }
             });
+
             db.setErrorHandler(new ErrorHandler() {
                 public void warning(SAXParseException e) throws SAXException {
                     LOG.warn("Warning parsing XML: {}", e.toString());
@@ -389,14 +398,14 @@ public class SiteMapParser {
         }
 
         // See if this is a sitemap index
-        NodeList nodeList = doc.getElementsByTagName("sitemapindex");
+        NodeList nodeList = doc.getElementsByTagNameNS("*", "sitemapindex");
         if (nodeList.getLength() > 0) {
-            nodeList = doc.getElementsByTagName("sitemap");
+            nodeList = doc.getElementsByTagNameNS("*", "sitemap");
             return parseSitemapIndex(sitemapUrl, nodeList);
-        } else if (doc.getElementsByTagName("urlset").getLength() > 0) {
+        } else if (doc.getElementsByTagNameNS("*", "urlset").getLength() > 0) {
             // This is a regular Sitemap
             return parseXmlSitemap(sitemapUrl, doc);
-        } else if (doc.getElementsByTagName("link").getLength() > 0) {
+        } else if (doc.getElementsByTagNameNS("*", "link").getLength() > 0) {
             // Could be RSS or Atom
             return parseSyndicationFormat(sitemapUrl, doc);
         }
@@ -436,7 +445,7 @@ public class SiteMapParser {
         SiteMap sitemap = new SiteMap(sitemapUrl);
         sitemap.setType(SitemapType.XML);
 
-        NodeList list = doc.getElementsByTagName("url");
+        NodeList list = doc.getElementsByTagNameNS(Namespace.SITEMAP, "url");
 
         // Loop through the <url>s
         for (int i = 0; i < list.getLength(); i++) {
@@ -444,10 +453,10 @@ public class SiteMapParser {
             Node n = list.item(i);
             if (n.getNodeType() == Node.ELEMENT_NODE) {
                 Element elem = (Element) n;
-                String lastMod = getElementValue(elem, "lastmod");
-                String changeFreq = getElementValue(elem, "changefreq");
-                String priority = getElementValue(elem, "priority");
-                String loc = getElementValue(elem, "loc");
+                String lastMod = getElementValue(Namespace.SITEMAP, elem, "lastmod");
+                String changeFreq = getElementValue(Namespace.SITEMAP, elem, "changefreq");
+                String priority = getElementValue(Namespace.SITEMAP, elem, "priority");
+                String loc = getElementValue(Namespace.SITEMAP, elem, "loc");
 
                 addUrlIntoSitemap(loc, sitemap, lastMod, changeFreq, priority, i);
             }
@@ -496,7 +505,7 @@ public class SiteMapParser {
 
             if (firstNode.getNodeType() == Node.ELEMENT_NODE) {
                 Element elem = (Element) firstNode;
-                String loc = getElementValue(elem, "loc");
+                String loc = getElementValue(Namespace.SITEMAP, elem, "loc");
 
                 // try the text content when no loc element
                 // has been specified
@@ -506,7 +515,7 @@ public class SiteMapParser {
 
                 try {
                     URL sitemapUrl = new URL(loc);
-                    String lastmod = getElementValue(elem, "lastmod");
+                    String lastmod = getElementValue(Namespace.SITEMAP, elem, "lastmod");
                     Date lastModified = SiteMap.convertToDate(lastmod);
 
                     // Right now we are not worried about sitemapUrls that point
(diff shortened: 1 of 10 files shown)


jnioche:
Thanks @grokwich, looks good at first glance but would need a more thorough review.

BTW you can apply the project's formatting rules with `mvn java-formatter:format`.

grokwich:
Cool.  Thx for the formatting tip.  Btw, if you wish to include the enhancement, feel free to hack up the code to suit your project.

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
```diff
@@ -436,18 +445,18 @@ protected SiteMap parseXmlSitemap(URL sitemapUrl, Document doc) {
         SiteMap sitemap = new SiteMap(sitemapUrl);
         sitemap.setType(SitemapType.XML);
 
-        NodeList list = doc.getElementsByTagName("url");
+        NodeList list = doc.getElementsByTagNameNS(Namespace.SITEMAP, "url");
```
jnioche: what would happen here if the namespace is not set explicitly in the XML? Why not do as above and just match based on the localname no matter what the NS is?

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
```diff
@@ -436,18 +445,18 @@ protected SiteMap parseXmlSitemap(URL sitemapUrl, Document doc) {
...
         for (int i = 0; i < list.getLength(); i++) {
 
             Node n = list.item(i);
             if (n.getNodeType() == Node.ELEMENT_NODE) {
                 Element elem = (Element) n;
-                String lastMod = getElementValue(elem, "lastmod");
-                String changeFreq = getElementValue(elem, "changefreq");
-                String priority = getElementValue(elem, "priority");
-                String loc = getElementValue(elem, "loc");
+                String lastMod = getElementValue(Namespace.SITEMAP, elem, "lastmod");
```
jnioche: same as above

src/main/java/crawlercommons/sitemaps/sax/DelegatorHandler.java
```diff
@@ -70,30 +72,33 @@ public void startElement(String uri, String localName, String qName, Attributes
...
     private void startRootElement(String uri, String localName, String qName, Attributes attributes) {
-        elementStack.push(qName);
-        if ("sitemapindex".equals(qName)) {
-            delegate = new XMLIndexHandler(url, elementStack, strict);
-        } else if ("urlset".equals(qName)) {
-            delegate = new XMLHandler(url, elementStack, strict);
-        } else if ("feed".equals(qName)) {
+        elementStack.push(localName);
+        
+        if ( Namespace.SITEMAP.equals(uri) ) {
```
jnioche: same here - why don't we consider only localnames regardless of the NS? What if the NS is not set?

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
grokwich: @jnioche i believe there would be no match without specifying the namespace.  Yes, there would be risk in breaking existing sitemaps that do not specify the namespace.  I was more forgiving about RSS/Atom, since they optionally allow namespaces.

But i believe the Sitemap XML spec requires a namespace...
https://www.sitemaps.org/protocol.html
https://support.google.com/webmasters/answer/183669#errors

And this validator fails when the namespace declaration is missing...
http://freetools.webmasterworld.com/tools/site-validator/

If you match any namespace, then you also risk incorrectly processing an element belonging to another namespace (example, i can imaging that <url> may exist in another mixed-in namespace).

So the decision seems to boil down to whether you wish to support invalid sitemap XML's that do not include the namespace decl.  If that is a common case, then i can see merit, but my impression is that google would reject a Sitemap XML without the ns?  

If you do go forward with this Pull and require the namespace, then it may be a good idea to fail-fast and report an error if the ns is missing?

sebastian-nagel: Hi @grokwich,

I've tried your pull request on a set of few 10,000 sitemaps using the SAX parser (see [1](https://groups.google.com/forum/#!topic/crawler-commons/pOLsCVwRsxY)). There is a significant impact: the number of extracted URLs drops from 193 million to 188 million. Afaics, that's (mostly) not caused by absent namespace declarations but by a namespace URL not equal to http://www.sitemaps.org/schemas/sitemap/0.9. E.g., in [this sitemap.xml](http://bvu.edu.vn/sitemap.xml):
```
<urlset xmlns="http://www.google.com/schemas/sitemap/0.9"
        xmlns:xhtml="http://www.w3.org/1999/xhtml">
```
It could be just an outdated location or version (like https://www.google.com/schemas/sitemap/0.84/). But is this really a reason to make the parser fail? What happens if a new sitemap version is released?

I agree that the parser should support namespaces but at least for the SAX parser robustness is most important, with the aim to extract URLs even if the sitemap is not valid. A validation mode or parser would be nice to have but it should be optional as it's just a different use case. If used for a crawler robustness and fault-tolerance matters.

grokwich: Hi @sebastian-nagel.

193M URL's is quite the test-bed:-)

Google Search Console has a sitemap testing tool.  When i test with no namespace decl i get a failure...
image

Also if i provide a bad namespace URL i get a failure...
image

But interestingly if i test using the URL examples you provided, then the test passes.  I guess this makes sense since the independent sitemap spec evolved from google's old internal sitemap format.  They don't seem to document those legacy namespaces, so i assume they are deprecated but still supported for legacy clients.

I also tried a few 3'rd party validators, and they reject the Google legacy URL's.

http://tools.seochat.com/tools/site-validator
http://freetools.webmasterworld.com/tools/site-validator

The spec is clear that the namespace is required...
https://www.sitemaps.org/protocol.html

Here is also an excerpt from a Google doc to troubleshoot failed processing, where they state that the namespace must exist and have the exact URL specified (expand the "Complete error list" at the bottom of the following page)...

https://support.google.com/webmasters/answer/183669#error-list

Here is the quote...
```
Unsupported format
Your sitemap is not in a supported format. Sitemaps must be in XML format, and use the correct header.
Common XML mistakes:

Your sitemap must use the correct header. For example, if your sitemap contains video information, it would have the following header:
<?xml version="1.0" encoding="UTF-8"?>
    <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"
        xmlns:video="http://www.google.com/schemas/sitemap-video/1.1">

The namespace in the header must be "http://www.sitemaps.org/schemas/sitemap/0.9" (not .9).
All XML attributes must enclosed in either single quotes (') or double quotes (") and those quotes must be straight, not curly. Word-processing programs such as Microsoft Word can insert curly quotes.
```
Another consideration is when a sitemap adds additional namespaces.  If one of those includes elements names that overlap with element names from sitemap XML, then you run the risk of incorrectly processing an element from another namespace.

For example, here is a case where you would incorrectly <foobar:url>...
```XML
<urlset 
    xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"
    xmlns:foobar="http://www.foobar.com/schemas/foobar">

   <url>
   ...
   </url>
   
   <foobar:url>
   ...
   </foobar:url>

   ....
<urlset> 
```
As for versioning, if a new namespace is introduced for sitemap (unlikely), then presumably new capability is being added that will change the way a client should process the doc.  In that case the sitemap parsers would have to be enhanced to support processing the new version, and clients should have to specify which version to parse with (default being the old version).

:warning: *All that being said*, i completely understand the concern of breaking existing (invalid) behavior.  One middle-ground would be to make the default wild-card, but allow the parser to be configured as strict (to avoid namespace clashes).

Whatever pragmatic decision you guys make is good with me.  Thx.

kkrugler: Hi @grokwich,

First off, many thanks for the very detailed response & testing! After dealing with all of the fun issues with invalid robot.txt files found in the wild, I appreciate your research and the time spent documenting things.

My two cents - I think your middle-ground solution is a good way to handle the two conflicting goals.

jnioche: +1 to the middle-ground approach
Thanks for the testing and commenting

jnioche:
@crawler-commons/committers @grokwich 
I have applied the formatting (although it looks a bit dodgy in places but that doesn't matter too much) + implemented the approach suggested above by allowing the namespaces to be anything when the strict flag is set to false. It lives in the branch[grokwich-add_namespace_aware](https://github.com/crawler-commons/crawler-commons/compare/grokwich-add_namespace_aware).

Could you please review it?

grokwich:
Hi @jnioche

The use of the wildcard namespace looks good to me.

Though I thought the "middle ground approach" was to match against the wildcard namespace by default (non-strict), and allow configuration that would require correct/strict use of the sitemap namespace?

If i'm reading correctly it appears that the default is "strict = true"?  I would personally be fine with that, but i thought the concern was breaking legacy cases that @sebastian-nagel identified?

You are also overloading Sitemap.strict.  The legacy meaning is related to sitemap baseurl processing, and now you are also including the correct use of the sitemap namespace (if strict == true ).

That may be fine, but it will constrain configuration to all strict (URL + namespace processing), or all non-strict.  It may also be the reason why you made namespace processing strict by default (since legacy URL processing is strict by default)? 

One other personal nice-to-have, if i configure for strict namespace processing it would be nice to confirm that the correct sitemap namespace is declared.  If not, then i would prefer to at least have a Warning logged (maybe even an Error)?   Currently the code will silently result in zero sitemap URL's if the the namespace is not correctly declared (and strict namespace processing is configured).

jnioche:
@grokwich thanks for your comments.

Yes, I have lumped together the control over the URLs and the namespace processing under the existing strict variable. Although semantically different they are relatively close in the sense that it is about enforcing the specifications. So you are right, by default we don't use the wildcard.

I am reluctant to change the API to add a separate boolean parameter for enforcing the namespace. Instead, we could have pass a Map<String,Object> around to hold the configuration. This would future-proof the code as we'd be able to add more config elements without breaking the API. However, this is a substantial change and I'd rather deal with it in a separate PR.

@sebastian-nagel in your benchmarks is there any way to see what we gained with this PR? There's bound to be sitemaps with namespaced elements which we could not get before. Also what do you think of the approach I described above i.e use wildcard when strict = false and add more subtle configuration separately?

@grokwich good point about logging, will add it

sebastian-nagel:
> what we gained

I've found only a single document in [sitemap-test-2017-03-04.warc.gz](s3://commoncrawl-seeds/sitemaps/test/sitemap-test-2017-03-04.warc.gz) `http://www.vanguardiapopular.org/sitemap04.xml`. Anyway, even it's only few, they're valid!

> lumped together the control over the URLs and the namespace processing under the existing strict variable. Although semantically different they are relatively close in the sense that it is about enforcing the specifications.

But the reasons to be strict are quite different: follow the format spec vs. avoid to be spammed. If #85 is implemented I would configure parser for commoncrawl.org to be strict and accept only URLs from the list of cross-submit hosts and the sitemap host itself. But parsing should be still permissive regarding format. The SAX parser already has a flag `allowPartial` to allow URLs from sitemaps only partially parsed due to XML errors or because they are truncated. I would opt for a new flag (`strictParsing` or `strictNamespace`).

jnioche:
Have opened a new PR #176

jnioche:
Implemented in #176 - thanks again @grokwich for your contribution and comments, much appreciated!

grokwich:
Sorry i was unable to review the latest pull.  Just took a peek, looks great.
I'm glad i could contribute @jnioche, thank you all for taking the time to push this into your nice lib.
Amazing regression analysis @sebastian-nagel!-)  Makes me even more confident using the lib as it continues to evolve.

PR #184, opened 2017-10-27
Use TLD for PLD code

kkrugler:
Trivial change to switch the PaidLevelDomain code to use EffectiveTldFinder

### DIFF
--- a/src/main/java/crawlercommons/domains/EffectiveTldFinder.java
+++ b/src/main/java/crawlercommons/domains/EffectiveTldFinder.java
@@ -82,12 +82,14 @@ import org.slf4j.LoggerFactory;
  */
 public class EffectiveTldFinder {
     private static final Logger LOGGER = LoggerFactory.getLogger(EffectiveTldFinder.class);
+    
     public static final String ETLD_DATA = "/effective_tld_names.dat";
     public static final String COMMENT = "//";
     public static final String DOT_REGEX = "\\.";
     public static final String EXCEPTION = "!";
     public static final String WILD_CARD = "*.";
     public static final char DOT = '.';
+    
     private static EffectiveTldFinder instance = null;
     private Map<String, EffectiveTLD> domains = null;
     private boolean configured = false;
@@ -222,7 +224,7 @@ public class EffectiveTldFinder {
 
     /**
      * This method uses the effective TLD to determine which component of a FQDN
-     * is the NIC-assigned domain name.
+     * is the NIC-assigned domain name (aka "Paid Level Domain").
      * 
      * @param hostname
      *            a string for which to obtain a NIC-assigned domain name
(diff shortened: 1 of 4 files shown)


src/main/java/crawlercommons/domains/PaidLevelDomain.java
```diff
@@ -64,60 +48,20 @@ public static String getPLD(String hostname) {
...
-                // Must be xxx.de.com format
-                firstHostPiece = numPieces - 3;
-            } else {
-                // Must be xxx.com format
-                firstHostPiece = numPieces - 2;
-            }
+        // Now use support in EffectiveTldFinder
+        String result = EffectiveTldFinder.getAssignedDomain(hostname, true);
+        if (result == null) {
+        	LOGGER.debug(String.format("Hostname %s isn't a valid FQDN", hostname));
```
jnioche: You don't need to use StringFormat with slf4j: just do 
`("Hostname {} isn't a valid FQDN", hostname)`
also the call to String.format is causing the build to fail, see https://travis-ci.org/crawler-commons/crawler-commons/builds/293871975#L1048

kkrugler: Hi @jnioche - Yeah, I've got projects with log4j and slf4j, I keep forgetting to use the slf4j formatting support. Also thanks for the heads-up re forbidden apis, I didn't realize I needed to run `mvn forbiddenapis:check` to catch these problems - I though `mvn package` did that.

kkrugler: Fixed

jnioche: Hi @kkrugler, I had forgotten about forbiddenapis but that shows that the automated checks with PRs are very useful. I'll have a closer look at the PR on Tuesday. Are you happy with me releasing 0.9 before merging it? we could release 0.9.1 shortly after that

kkrugler: Yes, please push out 0.9 - this is a very low priority change, just something I noticed in some code using the PLD support.

src/main/java/crawlercommons/domains/EffectiveTldFinder.java
```diff
@@ -308,13 +310,14 @@ private static String join(String[] ary) {
         public EffectiveTLD(String line) {
             if (line.startsWith(EXCEPTION)) {
                 exception = true;
-                domain = line.substring(1, line.length());
+                domain = line.substring(EXCEPTION.length(), line.length());
```
sebastian-nagel: +1

kkrugler:
Hi @sebastian-nagel - you're right, the PLD code should explicitly exclude private domains. Would it work for me to commit this, and then you make that change as part of your pull request #185?

sebastian-nagel:
Addressed @kkrugler's comment in #185: however, this requires a change in unit tests: `de.com` is part of the private section in the public suffixes list, hence `xxx.de.com` isn't a pay-level domain.

PR #186, opened 2017-11-05
Allow to include/exclude private domains / effective TLDs without re-initialization, fixes #185

### LINKED ISSUE #185: EffectiveTldFinder: allow to check whether a domain or public suffix is private without reinitialization
#172 introduced the option not to use public suffixes from the private section (including "blogspot.com" or "s3.cn-north-1.amazonaws.com.cn") to determine the domain name. The option is used while (re)reading the public suffix list. Because EffectiveTldFinder is a singleton it's not possible to check for both notions of "domain" in parallel. E.g., that's now done by Guava v23.3: [topDomainUnderRegistrySuffix()](http://google.github.io/guava/releases/23.3-jre/api/docs/com/google/common/net/InternetDomainName.html#topDomainUnderRegistrySuffix--) vs. [topPrivateDomain()](http://google.github.io/guava/releases/23.3-jre/api/docs/com/google/common/net/InternetDomainName.html#topPrivateDomain--).

Since the public suffix list is quite large (over 8,000 suffixes, 1,000 in the private section), the best way is to mark suffixes (represented by instances of the class EffectiveTld) whether they are private or not. This allows to add an additional boolean argument to all methods extracting domain or TLD whether to accept private domains/TLDs. The default should be `true` as currently.

sebastian-nagel:
- add optional flag excludePrivate to getAssignedDomain(...) / getEffectiveTLD(...)
- remove methods which exclude private eTLD during initialization

### DIFF
--- a/src/main/java/crawlercommons/domains/EffectiveTldFinder.java
+++ b/src/main/java/crawlercommons/domains/EffectiveTldFinder.java
@@ -82,14 +82,14 @@ import org.slf4j.LoggerFactory;
  */
 public class EffectiveTldFinder {
     private static final Logger LOGGER = LoggerFactory.getLogger(EffectiveTldFinder.class);
-    
+
     public static final String ETLD_DATA = "/effective_tld_names.dat";
     public static final String COMMENT = "//";
     public static final String DOT_REGEX = "\\.";
     public static final String EXCEPTION = "!";
     public static final String WILD_CARD = "*.";
     public static final char DOT = '.';
-    
+
     private static EffectiveTldFinder instance = null;
     private Map<String, EffectiveTLD> domains = null;
     private boolean configured = false;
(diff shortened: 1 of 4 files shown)


src/main/java/crawlercommons/domains/EffectiveTldFinder.java
```diff
@@ -195,8 +165,26 @@ public boolean initialize(InputStream effectiveTldDataStream, boolean excludePri
...
+     * @param excludePrivate
+     *            do not return an effective TLD from the PRIVATE section,
+     *            instead return the shorter eTLD not in the PRIVATE section
+     * @return the {@link EffectiveTLD}
+     */
+    public static EffectiveTLD getEffectiveTLD(String hostname, boolean excludePrivate) {
         if (getInstance().domains.containsKey(hostname)) {
-            return getInstance().domains.get(hostname);
+            EffectiveTLD foundTld = getInstance().domains.get(hostname);
+            if (!excludePrivate || !foundTld.isPrivate) {
```
kkrugler: Very minor nit, but !(excludePrivate && foundTld.isPrivate) is clearer to me.

PR #187, opened 2017-12-01
Add main() to EffectiveTldFinder

sebastian-nagel:
adds a main() method to EffectiveTldFinder
- for bulk testing
- or to extract domain names or public suffixes from a list of host names


sebastian-nagel:
Thanks, @jnioche for the review!

PR #190, opened 2017-12-15
Add SiteMapParser#walkSiteMap(URL,Consumer)

lucboruta:
This PR adds a convenience method to fetch a sitemap from the specified URL, recursively traverse any enclosed sitemap index, and perform the specified action for each sitemap URL. It is strongly inspired by methods like [`Iterable#forEach`](https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html#forEach-java.util.function.Consumer-) and [`Stream#forEach`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#forEach-java.util.function.Consumer-).

The code is based on [SitemapParser.java](https://github.com/Chaiavi/SitemapParser/blob/master/src/main/java/org/chaiware/SitemapParser.java) by @chaiavi, with extra checks to avoid NullPointerExceptions.

I added the method to `SiteMapParser` rather than `AbstractSiteMap` since the parser is needed to recurse into sitemap indices.

I'm open to suggestions regarding the method's name and null-friendliness.

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
+++ b/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
@@ -289,7 +289,8 @@ public class SiteMapParser {
      *             {@link java.net.URL}
      */
     public void walkSiteMap(URL onlineSitemapUrl, Consumer<SiteMapURL> action) throws UnknownFormatException, IOException {
-        if (action == null) {
+        if (onlineSitemapUrl == null || action == null) {
+            LOG.debug("Got null sitemap URL and/or action, stopping traversal");
             return;
         }
         walkSiteMap(parseSiteMap(onlineSitemapUrl), action);
@@ -318,6 +319,7 @@ public class SiteMapParser {
      */
     public void walkSiteMap(AbstractSiteMap sitemap, Consumer<SiteMapURL> action) throws UnknownFormatException, IOException {
         if (sitemap == null || action == null) {
+            LOG.debug("Got null sitemap and/or action, stopping traversal");
             return;
         }
         if (sitemap.isIndex()) {
--- a/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
+++ b/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
@@ -268,14 +268,14 @@ public class SiteMapParser {
     }
 
     /**
-     * Fetch a sitemap from the specified URL, recursively traverse any
-     * enclosed sitemap index, and perform the specified action for each
-     * sitemap URL until all URLs have been processed or the action throws an
-     * exception.
-     *
+     * Fetch a sitemap from the specified URL, recursively fetching and
+     * traversing the content of any enclosed sitemap index, and performing the
+     * specified action for each sitemap URL until all URLs have been processed
+     * or the action throws an exception.
+     * <p>
      * This method is a convenience method for a user who has a sitemap URL and
-     * wants a "Keep it simple" way to traverse it.
-     *
+     * wants a simple way to traverse it.
+     * <p>
      * Exceptions thrown by the action are relayed to the caller.
      *
      * @param onlineSitemapUrl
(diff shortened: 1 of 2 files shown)


lewismc:
@lucboruta this is really nice. I think a unit test to demonstrate a Consumer action would be ideal.

lucboruta:
Cool. Since we don't want unit tests to depend on remote content, I'll overload the first method with a few other methods working on byte arrays, content types, etc., as was done for `SiteMapParser#parseSiteMap`.

lucboruta:
Since the different `parseSiteMap` methods all return an `AbstractSiteMap`, duplicating all these signatures would be cumbersome. I simply added `SiteMapParser#walkSiteMap(AbstractSiteMap,Consumer)` and a unit test for that new method.

This PR now adds two public convenience methods:
* `SiteMapParser#walkSiteMap(URL,Consumer)` to fetch and traverse a sitemap, starting from the URL of the online sitemap;
* `SiteMapParser#walkSiteMap(AbstractSiteMap,Consumer)` to traverse a sitemap, starting from any `AbstractSiteMap`.

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
```diff
@@ -265,6 +267,78 @@ public AbstractSiteMap parseSiteMap(String contentType, byte[] content, URL url)
...
+            final Collection<AbstractSiteMap> links = ((SiteMapIndex) sitemap).getSitemaps();
+            for (final AbstractSiteMap asm : links) {
+                if (asm == null) {
+                    continue;
+                }
+                walkSiteMap(asm.getUrl(), action);
+            }
+        } else {
+            final Collection<SiteMapURL> links = ((SiteMap) sitemap).getSiteMapUrls();
+            for (final SiteMapURL url : links) {
```
sebastian-nagel: Just curious: why not write the loop as `links.stream().filter(Objects::nonNull).forEach(action);`?

sebastian-nagel:
+1 lgtm
- adds a useful API method (esp. by fetching recursively sitemaps of a sitemap index)
- for the long therm we could move the execution of the action to the parsing code to reduce the memory footprint

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
lucboruta: No good reason, it just mirrors the branch for sitemap indices above. I can add a commit with your version if you want.

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
```diff
@@ -265,6 +267,78 @@ public AbstractSiteMap parseSiteMap(String contentType, byte[] content, URL url)
...
+     *            The action to be performed for each element
+     * @throws UnknownFormatException
+     *             if there is an error parsing the sitemap
+     * @throws IOException
+     *             if there is an error fetching the content of any
+     *             {@link java.net.URL}
+     */
+    public void walkSiteMap(URL onlineSitemapUrl, Consumer<SiteMapURL> action) throws UnknownFormatException, IOException {
+        if (action == null) {
+            return;
```
Chaiavi: A comment using an slf4j logger will be appreciated here

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
```diff
@@ -265,6 +267,78 @@ public AbstractSiteMap parseSiteMap(String contentType, byte[] content, URL url)
...
+     *            The action to be performed for each element
+     * @throws UnknownFormatException
+     *             if there is an error parsing the sitemap
+     * @throws IOException
+     *             if there is an error fetching the content of any
+     *             {@link java.net.URL}
+     */
+    public void walkSiteMap(AbstractSiteMap sitemap, Consumer<SiteMapURL> action) throws UnknownFormatException, IOException {
+        if (sitemap == null || action == null) {
+            return;
```
Chaiavi: A comment using an slf4j logger will be appreciated here

lucboruta: Done.

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
lucboruta: Done.

I would usually throw a `NullPointerException` in this context, but I noticed that methods like `SiteMapParser#parseSiteMap(URL)` are null-safe, hence the early return statement, for the sake of consistency.

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
sebastian-nagel: No problem. It's a matter of taste whether to use "classic" loops or functional streams.

sebastian-nagel:
Resolved conflicts, rebased, updated changelog and merged 8ca46ff. Thanks, @lucboruta! Thanks, for the reviews (@Chaiavi, @lewismc)!

PR #198, opened 2018-03-27
Remove Tika dependency

kkrugler:
Quick attempt at replacing Tika-based mime-type detection with simple code.

I assume that sitemaps are UTF-8, thus no need to handle UTF-16LE/BE encoded BOMs when detecting patterns.

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
+++ b/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
@@ -17,9 +17,8 @@
 package crawlercommons.sitemaps;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static org.apache.tika.mime.MediaType.APPLICATION_XML;
-import static org.apache.tika.mime.MediaType.TEXT_PLAIN;
 
+import java.io.BufferedInputStream;
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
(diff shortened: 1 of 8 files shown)


src/main/java/crawlercommons/sitemaps/SiteMapParser.java
```diff
@@ -228,41 +214,28 @@ public AbstractSiteMap parseSiteMap(byte[] content, URL url) throws UnknownForma
...
-                    // re-open decompressed stream and parse as text
-                    decompressed = new GZIPInputStream(new ByteArrayInputStream(content));
+                } else if (mimeTypeDetector.isText(compressedType)) {
                     return processText(url, decompressed);
-                } else if (GZ_MEDIA_TYPES.contains(embeddedType)) {
-                    throw new UnknownFormatException("Can't parse gzip recursively: " + url);
                 }
-                throw new UnknownFormatException("Can't parse a gzipped sitemap with the embedded MediaType of: " + embeddedType + " (at: " + url + ")");
+            } catch (Exception e) {
+                String msg = String.format("Failed to detect embedded MediaType of gzipped sitemap '%s'", url);
```
jnioche: use explicit locale?  
The forbidden API complains otherwise
```
[ERROR] Forbidden method invocation: java.lang.String#format(java.lang.String,java.lang.Object[]) [Uses default locale]
[ERROR]   in crawlercommons.sitemaps.SiteMapParser (SiteMapParser.java:232)
[ERROR] Forbidden method invocation: java.lang.String#format(java.lang.String,java.lang.Object[]) [Uses default locale]
[ERROR]   in crawlercommons.sitemaps.SiteMapParser (SiteMapParser.java:237)
```

sebastian-nagel:
Sorry, you can call me a fool but I've run the large tests on the wrong version of crawler-commons (didn't update the dependency to 0.10-SNAPSHOT). There are a couple of regressions, but nothing serious. Opened #200 to fix these.

PR #200, opened 2018-04-10
Improve MIME detection for sitemaps

sebastian-nagel:
- avoid NPE if no MIME type has been detected
- allow optional leading white space before MIME patterns (after optional BOM)

### DIFF
--- a/src/main/java/crawlercommons/mimetypes/MimeTypeDetector.java
+++ b/src/main/java/crawlercommons/mimetypes/MimeTypeDetector.java
@@ -44,25 +44,33 @@ public class MimeTypeDetector {
                     (byte) 0xBF
                     };
 
+    private static final int LEADING_WHITESPACE_MAX_SKIP = 16;
+
+    private final static boolean[] spaceCharacters = new boolean[256];
+    static {
+        spaceCharacters[0x09] = true; // \t - character tabulation (ht)
+        spaceCharacters[0x0a] = true; // \n - line feed (lf)
+        spaceCharacters[0x0b] = true; // line tabulation (vt)
+        spaceCharacters[0x0c] = true; // form feed (ff)
+        spaceCharacters[0x0d] = true; // \r - carriage return (cr)
+        spaceCharacters[0x20] = true; // space
+    }
+
     private static class MimeTypeEntry {
         private String mimeType;
         private byte[] pattern;
+        private boolean allowBOM;
+        private boolean allowLeadingSpace;
 
         public MimeTypeEntry(String mimeType, String pattern) {
-            this(mimeType, pattern, false);
+            this(mimeType, pattern, false, false);
         }
 
-        public MimeTypeEntry(String mimeType, String pattern, boolean addBOM) {
+        public MimeTypeEntry(String mimeType, String pattern, boolean allowBOM, boolean allowLeadingSpace) {
             this.mimeType = mimeType;
-
-            byte[] patternBytes = pattern.getBytes(StandardCharsets.UTF_8);
-            if (addBOM) {
-                this.pattern = new byte[UTF8_BOM.length + patternBytes.length];
-                System.arraycopy(UTF8_BOM, 0, this.pattern, 0, UTF8_BOM.length);
-                System.arraycopy(patternBytes, 0, this.pattern, UTF8_BOM.length, patternBytes.length);
-            } else {
-                this.pattern = patternBytes;
-            }
+            this.allowBOM = allowBOM;
+            this.allowLeadingSpace = allowLeadingSpace;
+            this.pattern = pattern.getBytes(StandardCharsets.UTF_8);
         }
 
         public MimeTypeEntry(String mimeType, int... pattern) {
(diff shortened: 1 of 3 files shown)


kkrugler:
Hi @sebastian-nagel - thanks for the PR. I took an initial look at the changes. The logic for handling leading whitespace seemed a bit overly complex at first glance, but I'd like to review more (pull the branch and look at it in Eclipse) before commenting.

sebastian-nagel:
- improved the log messages for gzip-embedded sitemaps. The message
```Can't parse a sitemap with MediaType 'application/gzip' from '...'```
now becomes
```Failed to detect embedded MediaType of gzipped sitemap '...'```
- add support for application/rdf+xml (wrapping RSS feed)

File: src/main/java/crawlercommons/mimetypes/MimeTypeDetector.java [hunk 1]
```diff
@@ -44,25 +45,33 @@
...
+        spaceCharacters[0x0b] = true; // line tabulation (vt)
+        spaceCharacters[0x0c] = true; // form feed (ff)
+        spaceCharacters[0x0d] = true; // \r - carriage return (cr)
+        spaceCharacters[0x20] = true; // space
+    }
+
     private static class MimeTypeEntry {
         private String mimeType;
         private byte[] pattern;
+        private boolean allowBOM;
```
kkrugler: I think just a "isText" flag, which implies both BOM and leading whitespace.

src/main/java/crawlercommons/mimetypes/MimeTypeDetector.java
```diff
@@ -95,43 +104,65 @@ public MimeTypeDetector() {
...
     }
 
     public String detect(byte[] content, int offset, int length) {
+        int offsetBOM = -1;
+        int offsetSpace = -1;
         for (MimeTypeEntry entry : mimeTypes) {
             if (patternMatches(entry.getPattern(), content, offset, length)) {
                 return entry.getMimeType();
             }
+            if (entry.allowBOM) {
```
kkrugler: BOM has to be at the beginning. I also think we should simplify and assume offset == 0 always (nobody calls detect with non-zero offset). So just have an isBOM check (which advances the offset by BOM length), and then an isWhitespace check (same thing) and then the previous logic.

src/main/java/crawlercommons/mimetypes/MimeTypeDetector.java
```diff
@@ -95,43 +104,65 @@ public MimeTypeDetector() {
...
+        mimeTypes.add(new MimeTypeEntry(TEXT_MIMETYPES[0], "http://", true, true));
+        mimeTypes.add(new MimeTypeEntry(TEXT_MIMETYPES[0], "https://", true, true));
 
         mimeTypes.add(new MimeTypeEntry(GZIP_MIMETYPES[0], "\037\213"));
         mimeTypes.add(new MimeTypeEntry(GZIP_MIMETYPES[0], 0x1F, 0x8B));
 
         maxPatternLength = 0;
         for (MimeTypeEntry entry : mimeTypes) {
-            maxPatternLength = Math.max(maxPatternLength, entry.getPattern().length);
+            int length = entry.getPattern().length;
```
kkrugler: As per comment below, I think we can skip this by just treating BOM and leading whitespace checks as special, and adjust the offset for those, then fall into the regular pattern check.

src/main/java/crawlercommons/mimetypes/MimeTypeDetector.java
sebastian-nagel: Agreed.

src/main/java/crawlercommons/mimetypes/MimeTypeDetector.java
sebastian-nagel: > BOM has to be at the beginning.

Of course, but in my test set there are sitemaps with two BOMs. We would need also to change the BOMInputStream (also to skip white space), but detection comes first.

> assume offset == 0 always (nobody calls detect with non-zero offset)

Agreed.

src/main/java/crawlercommons/mimetypes/MimeTypeDetector.java
sebastian-nagel: The method `detect(InputStream is)` requires the max. pattern length to create the byte[] array. The length must be longer for "text" patterns. But with a isTextPattern flag, this could be simplified by just adding LEADING_WHITESPACE_MAX_SKIP.

sebastian-nagel:
Thanks, @kkrugler. I'll update the PR tomorrow.

src/main/java/crawlercommons/mimetypes/MimeTypeDetector.java
```diff
@@ -104,62 +102,53 @@ public MimeTypeDetector() {
...
     public String detect(byte[] content) {
-        return detect(content, 0, content.length);
+        return detect(content, content.length);
     }
 
-    public String detect(byte[] content, int offset, int length) {
-        int offsetBOM = -1;
-        int offsetSpace = -1;
+    public String detect(byte[] content, int length) {
+        int offsetText = -1;
```
kkrugler: I don't understand why this is initialized here, and changed in the per-mimetype loop below (more comments there).

kkrugler: OK, I get it; you only want to do the advance once.

Issue #196, opened 2018-03-18
Add JAX-B dependencies to POM


YossiTamari:
From Java 9, JAX-B is not included in the JSE.
AbstractSiteMap imports javax.xml.bind.DatatypeConverter.
There is a workaround parameter for Java 9, but the correct solution is to add dependencies.
See also: https://docs.oracle.com/javase/9/migrate/#GUID-F640FA9D-FB66-4D85-AD2B-D931174C09A3, https://stackoverflow.com/a/43574427/9141985

jnioche:
Thakns @YossiTamari would you like to submit a PR?

YossiTamari:
Too much work... I just used `--add-modules java.xml.bind` locally. The lines that need to be added in the POM are in the SO answer I mentioned.

jnioche:
It would probably be better to avoid the use of this class altogether and rely on the core Java SE code. That would also be one less dependency that we'd have to deal with. 

I have seen this `java.time.format.DateFormatter.ISO_OFFSET_DATE_TIME ` being used as a replacement in [this patch](https://github.com/mongodb/mongo-java-driver/commit/307d502c67256216ae2e0e3d54d0e4e5cea6ba1b)

YossiTamari:
Agreed. It will probably also simplify the code, by removing the W3C_NO_SECONDS_PATTERN section.

PR #207, opened 2018-05-02
Add JAX-B dependencies to POM

jnioche:
see #196 
java.time.format.DateFormatter.ISO_OFFSET_DATE_TIME can't be used as a straightforward replacement  so instead I opted to add the dependency


jnioche:
thanks @kkrugler

PR #211, opened 2018-05-15
Allow for legacy URIs when checking sitemap namespaces

sebastian-nagel:
- e.g., allow legacy namespace URI but ignore URLs
  from image and video sitemap extensions
- resolve relative namespace URIs
- add namespace URIs of sitemap extensions (news, images, videos)

E.g., to check the namespace but allow legacy URIs and also sitemaps without a namespace declaration:
```
sitemapParser.setStrictNamespace(true);
sitemapParser.addAcceptedNamespace(Namespace.SITEMAP_LEGACY);
sitemapParser.addAcceptedNamespace(Namespace.EMPTY);
```

To allow image URLs one could add
```
sitemapParser.addAcceptedNamespace(Namespace.IMAGE);
```
It's also easy to add any other namespace URI.

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
+++ b/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
@@ -28,7 +28,9 @@ import java.io.StringReader;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Locale;
+import java.util.Set;
 import java.util.function.Consumer;
 import java.util.zip.GZIPInputStream;
 
@@ -79,6 +81,9 @@ public class SiteMapParser {
      **/
     protected boolean strictNamespace = false;
 
+    /** Set of namespaces (if {@link #strictNamespace}) accepted by the parser. URLs from other namespaces are ignored. */
+    protected Set<String> acceptedNamespaces = new HashSet<>();
+
     private MimeTypeDetector mimeTypeDetector;
 
     public SiteMapParser() {
@@ -106,7 +111,8 @@ public class SiteMapParser {
 
     /**
      * @return whether the parser allows any namespace or just the one from the
-     *         specification
+     *         specification (or any namespace accepted,
+     *         {@link #addAcceptedNamespace(String)})
      */
     public boolean isStrictNamespace() {
         return strictNamespace;
@@ -114,10 +120,34 @@ public class SiteMapParser {
 
     /**
      * Sets the parser to allow any namespace or just the one from the
-     * specification
+     * specification (or any namespace accepted,
+     * {@link #addAcceptedNamespace(String)})
      */
     public void setStrictNamespace(boolean s) {
         strictNamespace = s;
+        if (strictNamespace) {
+            addAcceptedNamespace(Namespace.SITEMAP);
+        }
+    }
+
+    /**
+     * Add namespace URI to set of accepted namespaces.
+     * 
+     * @param namespaceUri
+     */
+    public void addAcceptedNamespace(String namespaceUri) {
+        acceptedNamespaces.add(namespaceUri);
+    }
+
+    /**
+     * Add namespace URIs to set of accepted namespaces.
+     * 
+     * @param namespaceUris
+     */
+    public void addAcceptedNamespace(String[] namespaceUris) {
+        for (String namespaceUri : namespaceUris) {
+            acceptedNamespaces.add(namespaceUri);
+        }
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/main/java/crawlercommons/sitemaps/SiteMapParser.java
```diff
@@ -106,18 +111,43 @@ public boolean isStrict() {
...
      * Sets the parser to allow any namespace or just the one from the
-     * specification
+     * specification (or any namespace accepted,
+     * {@link #addAcceptedNamespace(String)})
      */
     public void setStrictNamespace(boolean s) {
         strictNamespace = s;
+        if (strictNamespace) {
+            addAcceptedNamespace(Namespace.SITEMAP);
+        }
```
kkrugler: I can see why you auto-add the default sitemap namespace, but side-effects to what looks like a straight-up POJO setter should be clearly called out. I'd suggest documenting that in the function Javadoc (`setting this to true will always add Namespace.SITEMAP`)

src/main/java/crawlercommons/sitemaps/sax/DelegatorHandler.java
```diff
@@ -107,19 +112,48 @@ private void startRootElement(String uri, String localName, String qName, Attrib
         // the <rss> or <rdf> tag is a <channel> tag, so we can use that.
         // See https://github.com/crawler-commons/crawler-commons/issues/87
         // and also RSS 1.0 specification http://web.resource.org/rss/1.0/spec
-        else if ("channel".equals(localName)) {
+        else if ("rss".equals(localName)) {
+            return; // ignore
```
kkrugler: Is this related to improving namespace support?

src/main/java/crawlercommons/sitemaps/sax/XMLIndexHandler.java
```diff
@@ -71,7 +70,7 @@ public void startElement(String uri, String localName, String qName, Attributes
     }
 
     public void endElement(String uri, String localName, String qName) throws SAXException {
-        if (isStrictNamespace() && !Namespace.SITEMAP.equals(uri)) {
+        if (isStrictNamespace() && !isAcceptedNamespace(uri)) {
```
kkrugler: `XMLHandler` also has this same check in its `startElement` method, but not in this `XMLIndexHandler` - wondering why we don't need it here.

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
sebastian-nagel: It was merely to keep the API unchanged. I'll update the Javadoc.

src/main/java/crawlercommons/sitemaps/sax/DelegatorHandler.java
sebastian-nagel: Ok, I remember: at an early stage I've added this to fix that also rss feeds are checked for namespace and then are always rejected because of the now removed else-if-condition three lines below:
```
Failed to parse sitemap http://www.globes.co.il/webservice/rss/rssfeeder.asmx/FeederNode?iID=2173: crawlercommons.sitemaps.UnknownFormatException: Namespace  does not match standard namespace http://www.sitemaps.org/schemas/sitemap/0.9
```
That's fixed anyway with this PR. Either I forgot about it or kept it to avoid the debug log message below: "Skipped unknown root element <rss> in ..."

I'll remove it and open a new issue to review skipping of "unknown" elements until the "root" element is found:
- it should also be done for <rdf>
- while for <rss> and <rdf> we need to step down into the tree and wait for <channel>, this shouldn't be done for other "unknown" tags. E.g., from the debug log when an XHTML document is parsed as sitemap (it's valid XML):
```
18/06/05 11:10:32 DEBUG sitemaps.SiteMapParser:124 - Skipped unknown root element <html> in ...
18/06/05 11:10:32 DEBUG sitemaps.SiteMapParser:124 - Skipped unknown root element <head> in ...
18/06/05 11:10:32 DEBUG sitemaps.SiteMapParser:124 - Skipped unknown root element <title> in ...
18/06/05 11:10:32 DEBUG sitemaps.SiteMapParser:124 - Skipped unknown root element <meta> in ...
```

src/main/java/crawlercommons/sitemaps/sax/XMLIndexHandler.java
sebastian-nagel: I think I've only replaced all occurrences of `!Namespace.SITEMAP.equals(uri)`. But:
- XMLIndexHandler.startElement(...) is currently a noop because other than in XMLHandler we do not accept URLs from elements not properly closed. Btw., that came recently up on the [Nutch mailing list](https://lists.apache.org/thread.html/8ebcffbe2bd8edafb6030e4f28fceee07aee08a1ce06a94755ee8d74@%3Cuser.nutch.apache.org%3E). If this is addressed, also the startElement method will 
- I haven't seen so far a sitemap index which uses two or more namespaces. For sitemaps that's quite common because of the many extensions (video, image, news) or embeded XHTML.

jnioche:
thanks @sebastian-nagel and @kkrugler

PR #216, opened 2018-08-01
Fix for URL with only a query string (no path)

kkrugler:
Fix for processing URLs which are just `domain.com?<query parameters>` (no path). These would get turned into a string that looks like `?<query parameters>`, instead of `/?<query parameters>`, which means they wouldn't match any rules (which all start with `/`).


kkrugler:
Thanks @sebastian-nagel - my Eclipse preferences seem off, I'll have to fix those. Formatting corrected.

sebastian-nagel:
Thanks, @kkrugler!

PR #217, opened 2018-09-20
Use the Java 8 date and time API (java.time.*) to parse dates in sitemaps

sebastian-nagel:
- use thread-safe DateTimeFormatter instead of ThreadLocal<DateFormat>
- simplify parsing of RSS publication dates
- remove obsolete regex pattern to catch dates with time zone but without seconds (covered by DateTimeFormatter.ISO_OFFSET_DATE_TIME)
- extend unit tests

Makes the code simpler and parsing of RSS feeds 10-20% faster by using a single date format instead of 4 formats tried after each other.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,7 @@
 Crawler-Commons Change Log
 
 Current Development 0.11-SNAPSHOT (yyyy-mm-dd)
+- [Sitemaps] Use the Java 8 date and time API (java.time.*) to parse dates in sitemaps (sebastian-nagel) #217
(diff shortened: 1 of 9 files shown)


jnioche:
Thanks @sebastian-nagel. The CI chokes on Javadoc issues which should be easy to fix, I'll look at your PR in more details shortly

sebastian-nagel:
Ok, the Javadoc error and further warnings are fixed. I've also seen that the fix for #196 is obsoleted by the removal of the import of `javax.xml.bind.DatatypeConverter`. Thanks, @jnioche!

jnioche:
Great, thanks @sebastian-nagel

PR #220, opened 2018-12-03
In case of the use of a different locale, price tag can be formatted …

goldenlink:
…with ',' instead of '.' leading to a NPE.

NPE is generated because parseFloat returns a Float object that can be set null in case of NumberFormatException, but the VideoPrice accepts only float.
To bypass this issue and avoid reccuring errors, I've moved the VideoPrice price field to a Float object instead accepting null in case of.

It is far from ideal, and parseFloat would enjoy being able to parse different locale formatting. Anyway, in a first quick fix, this allows the rest of the file to be parsed,
whereas the previous error had all the file to fail while parsing.

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/extension/VideoAttributes.java
+++ b/src/main/java/crawlercommons/sitemaps/extension/VideoAttributes.java
@@ -456,7 +456,7 @@ public class VideoAttributes extends ExtensionMetadata {
             }
             VideoPrice that = (VideoPrice) other;
             return Objects.equals(currency, that.currency) //
-                            && price == that.price //
+                            && Objects.equals(price, that.price) //
                             && type == that.type //
                             && Objects.equals(resolution, that.resolution);
         }
@@ -477,7 +477,7 @@ public class VideoAttributes extends ExtensionMetadata {
             return price;
         }
 
-        public void setPrice(float price) {
+        public void setPrice(Float price) {
             this.price = price;
         }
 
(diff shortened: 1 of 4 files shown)


kkrugler:
This is a currency value. Instead of trying to parse it as a float, it needs special handling. Lots of articles online about this, e.g. https://dzone.com/articles/currency-format-validation-and. If the locale is unknown, then best is a (limited) set of common locales that are tried until it parses successfully. If it fails, then logging the error and returning something that is ultimately interpreted as 0.0 might be a better option, versus a null Float object.

sebastian-nagel:
After a closer look into the [video sitemap schema](https://www.google.com/schemas/sitemap-video/1.1/sitemap-video.xsd): the price is of type "xsd:string" and "The price value must either be a non-negative  [decimal](http://www.datypic.com/sc/xsd/t-xsd_decimal.html) or be empty." Empty values are now ignored, see #221.
Of course, we may want to improve the parser to process even localized numbers. But I would accept this PR as a first step: make the price value optional probably and do not fail with a NPE if the value fails to parse as a float. Any thoughts on this? @goldenlink, would you be able to extend the PR to parse localized prices?

goldenlink:
Hi, well, that was my trade-off too. I wanted it to be non blocking, and avoid failure, even if missing some data. I'll try to make a localized currency instead when I'll find the time for it.

sebastian-nagel:
Hi @goldenlink, together with a fix for #221 I've "inlined" the unit test to avoid that every test adds its own sitemap file. Hope this is ok.

PR #223, opened 2018-12-07
Allow empty price in video sitemaps, fixes #221

### LINKED ISSUE #221: Allow empty price in video sitemaps
The [video sitemap schema](https://www.google.com/schemas/sitemap-video/1.1/sitemap-video.xsd) allows "the price value ... be empty." With an empty price, it's possible to express whether the video is for rent or purchase, e.g. `<video:price type="purchase" />`. This information should be added to the video attributes, as of now <video:price> elements with empty value are ignored. See also #220 which makes the price value optional in case the number cannot be parsed.

sebastian-nagel:
Fix for #221 including unit test.
It's based on and includes PR #220.


src/main/java/crawlercommons/sitemaps/extension/VideoAttributes.java
```diff
@@ -424,17 +424,17 @@ public void setLive(Boolean live) {
         /**
          * Video price
          */
-        private float price;
+        private Float price;
 
-        public VideoPrice(final String currency, final float price) {
```
kkrugler: Not a comment on your change, but this use of `final` in method parameters is...not very useful or standard :)

src/test/java/crawlercommons/sitemaps/SiteMapParserExtensionTest.java
```diff
@@ -74,12 +78,25 @@ public void testVideosSitemap() throws UnknownFormatException, IOException {
...
 
-        for (SiteMapURL su : sm.getSiteMapUrls()) {
-            assertNotNull(su.getAttributesForExtension(Extension.VIDEO));
-            VideoAttributes attr = (VideoAttributes) su.getAttributesForExtension(Extension.VIDEO)[0];
-            assertEquals(expectedVideoAttributes, attr);
-        }
+        // locale-specific number format in <video:price>, test #220
+        expectedVideoAttributes = new VideoAttributes(new URL("http://www.example.com/thumbs/123-2.jpg"), "Grilling steaks for summer, episode 2",
+                        "Alkis shows you how to get perfectly done steaks every time", new URL("http://www.example.com/video123-2.flv"), null);
+        expectedVideoAttributes.setPrices(new VideoAttributes.VideoPrice[] { new VideoAttributes.VideoPrice("EUR", null, VideoAttributes.VideoPriceType.own) });
```
kkrugler: Maybe a comment about why we're setting the price to null (current expected behavior is to not handle non-US locale currency values, but set them to null vs. failing).

src/main/java/crawlercommons/sitemaps/sax/extension/VideoHandler.java
```diff
@@ -101,7 +101,36 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
         if (currAttr == null) {
             currAttr = new VideoAttributes();
         }
-        if (value.isEmpty()) {
+        if ("price".equals(localName)) {
```
kkrugler: Looks like all this code is just moving from another location, so I'm not actually reviewing it.

PR #232, opened 2019-02-20
EffectiveTldFinder throws IllegalArgumentException on IDN domain names containing prohibited characters, fixes #231

### LINKED ISSUE #231: EffectiveTldFinder throws IllegalArgumentException on IDN domain names containing prohibited characters
EffectiveTldFinder throws an IllegalArgumentException when parsing domain names containing characters prohibited in [Internationalized Domain Names](https://en.wikipedia.org/wiki/Internationalized_domain_name) (IDNs):
```
Exception in thread "main" java.lang.IllegalArgumentException: java.text.ParseException: A prohibited code point was found in the inputp?
        at java.net.IDN.toASCIIInternal(IDN.java:274)
        at java.net.IDN.toASCII(IDN.java:122)
        at java.net.IDN.toASCII(IDN.java:151)
        at crawlercommons.domains.EffectiveTldFinder$EffectiveTLD.asciiConvert(EffectiveTldFinder.java:480)
        at crawlercommons.domains.EffectiveTldFinder$EffectiveTLD.normalizeName(EffectiveTldFinder.java:428)
        at crawlercommons.domains.EffectiveTldFinder$EffectiveTLD.<init>(EffectiveTldFinder.java:416)
        at crawlercommons.domains.EffectiveTldFinder.findEffectiveTLD(EffectiveTldFinder.java:235)
        at crawlercommons.domains.EffectiveTldFinder.getAssignedDomain(EffectiveTldFinder.java:289)
        at crawlercommons.domains.EffectiveTldFinder.main(EffectiveTldFinder.java:391)
Caused by: java.text.ParseException: A prohibited code point was found in the inputp?
        at sun.net.idn.StringPrep.prepare(StringPrep.java:450)
        at java.net.IDN.toASCIIInternal(IDN.java:272)
        ... 8 more
```

Specifically, the exception was caused by a wrong encoding conversion which caused that the domain name contained the replacement character "�" (U+FFFD) which is prohibited by [RFC3490](https://tools.ietf.org/html/rfc3490) resp. [RFC3454](https://tools.ietf.org/html/rfc3454#page-78). The exception only occurs if the prohibited character is contained in the dot-separated segment "matched" by a wildcard public suffix (e.g., `*.kawasaki.jp`). 

A fix should address these two points:
1. catch the runtime exception and return null or the host name depending on the configuration (parameter `strict`). Alternatively, declare and document it to avoid that applications unexpectedly crash.
2. define the behavior of EffectiveTldFinder in case the domain or host name contains prohibited characters and there is no valid ASCII representation.

sebastian-nagel:
- catch IllegalArgumentException when converting IDNs to ASCII
- validate domain names returned by getAssignedDomain(...)
- add unit tests, complete Javadoc

The behavior of `EffectiveTldFinder.getAssignedDomain(hostname)` in case there are prohibited characters is now (see updated unit test):
1. in the public suffix ("effective TLD") or the domain name: return null or hostname depending on the `strict` flag
2. as part of the hostname but not the domain name part: return the valid domain name

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,7 @@
 Crawler-Commons Change Log
 
 Current Development 0.11-SNAPSHOT (yyyy-mm-dd)
+- EffectiveTldFinder throws IllegalArgumentException on IDN domain names containing prohibited characters (sebastian-nagel) #231
 - [Sitemaps] Sitemap index: stop URL at closing </loc> (sebastian-nagel, kkrugler) #213
 - [Sitemaps] Allow empty price in video sitemaps (sebastian-nagel) #221
 - [Sitemaps] In case of the use of a different locale, price tag can be formatted with ',' instead of '.' leading to a NPE (goldenlink) #220
--- a/src/main/java/crawlercommons/domains/EffectiveTldFinder.java
+++ b/src/main/java/crawlercommons/domains/EffectiveTldFinder.java
@@ -222,6 +222,7 @@ public class EffectiveTldFinder {
                 if (offset == 0 || foundTld.isException() || !foundTld.isWild()) {
                     return res;
                 }
+
                 // wildcards create an open ETLD namespace
                 int wildcardOffset = hostname.lastIndexOf(DOT, offset - 2);
                 String retryTld;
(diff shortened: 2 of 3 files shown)


sebastian-nagel:
Thanks, @kkrugler. `java.net.IDN.toAscii(null)` throws a NullPointerException.

Issue #226, opened 2018-12-12
XMLHandler needs to append text in characters() vs. immediately processing


kkrugler:
Currently the code in `XMLHandler.characters()` is:

``` java
        if ("loc".equals(localName) || "url".equals(localName)) {
            loc.append(value);
        } else if ("changefreq".equals(localName)) {
            changeFreq = value;
        } else if ("lastmod".equals(localName)) {
            lastMod = value;
        } else if ("priority".equals(localName)) {
            priority = value;
        }
```

So only `<loc>` elements get appended, others are processed immediately.

Really we should have a single text collector that is appended to in `characters()`, and it's used in the `endElement()` method (and then cleared).

PR #233, opened 2019-02-22
Buffer characters when characters(...) is called, fixes #225 and #226

### LINKED ISSUE #225: XMLIndexHandler needs to accumulate the lastmod date string before parsing
Currently the code in `XMLIndexHandler.characters()` just tries to parse whatever it's passed as a date. But this handler can (and will) be passed text incrementally, so it needs to be accumulated, and in the `endElement()` method it can then (a) trim the string, and (b) try to parse the date.

You can see this behavior if you use a `<![CDATA[xxx]]>` format for the `<lastmod>` data, where `characters()` is called with some spaces and a newline, before getting called with the actual string.

sebastian-nagel:
Buffer characters when characters(...) is called, instead of immediately processing the current character chunk:
- fixes errors (#225 and #226) when character chunks are interrupted by CDATA sections or character entities
- DelegatorHandler provides a character buffer, so that derived classes can append characters to it and finally get the buffered content
- includes code cleanup in all handler classes:
  - add `@Override` annotations
  - remove stubb method implementations

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/sax/DelegatorHandler.java
+++ b/src/main/java/crawlercommons/sitemaps/sax/DelegatorHandler.java
@@ -48,6 +48,7 @@ public class DelegatorHandler extends DefaultHandler {
     private UnknownFormatException exception;
     private Set<String> acceptedNamespaces;
     protected Map<String, Extension> extensionNamespaces;
+    protected StringBuilder characterBuffer = new StringBuilder();
 
     protected DelegatorHandler(LinkedList<String> elementStack, boolean strict) {
         this.elementStack = elementStack;
@@ -103,6 +104,7 @@ public class DelegatorHandler extends DefaultHandler {
         return exception;
     }
 
+    @Override
     public void startElement(String uri, String localName, String qName, Attributes attributes) throws SAXException {
         if (elementStack.isEmpty() || delegate == null) {
             startRootElement(uri, localName, qName, attributes);
@@ -170,6 +172,7 @@ public class DelegatorHandler extends DefaultHandler {
         delegate.setExtensionNamespaces(extensionNamespaces);
     }
 
+    @Override
     public void endElement(String uri, String localName, String qName) throws SAXException {
         if (delegate != null) {
             delegate.endElement(uri, localName, qName);
@@ -177,10 +180,36 @@ public class DelegatorHandler extends DefaultHandler {
         elementStack.pop();
     }
 
+    @Override
     public void characters(char ch[], int start, int length) throws SAXException {
         if (delegate != null) {
             delegate.characters(ch, start, length);
         }
+    }
+
+    protected void appendCharacterBuffer(char ch[], int start, int length) {
+        if (characterBuffer != null) {
+            for (int i = start; i < start + length; i++) {
+                characterBuffer.append(ch[i]);
+            }
+        }
+    }
+
+    protected void appendCharacterBuffer(String str) {
+        characterBuffer.append(str);
+    }
+
+    protected String getAndResetCharacterBuffer() {
+        if (characterBuffer == null) {
+            return null;
+        }
+        String value = characterBuffer.toString();
+        resetCharacterBuffer();
+        return value;
+    }
+
+    protected void resetCharacterBuffer() {
+        characterBuffer = new StringBuilder();
     }
 
     protected String currentElement() {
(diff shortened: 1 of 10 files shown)


src/main/java/crawlercommons/sitemaps/sax/DelegatorHandler.java
```diff
@@ -170,19 +172,46 @@ else if ("channel".equals(localName)) {
...
 
+    @Override
     public void characters(char ch[], int start, int length) throws SAXException {
         if (delegate != null) {
             delegate.characters(ch, start, length);
         }
     }
 
+    protected void appendCharacterBuffer(char ch[], int start, int length) {
+        if (characterBuffer != null) {
```
kkrugler: When would `characterBuffer` ever be null? Same question for `getAndResetCharacterBuffer()`

src/main/java/crawlercommons/sitemaps/sax/RSSHandler.java
```diff
@@ -109,47 +107,55 @@ public void startElement(String uri, String localName, String qName, Attributes
      */
     @Override
     public void endElement(String uri, String localName, String qName) throws SAXException {
-        if ("link".equals(currentElement()) || "guid".equals(currentElement())) {
+        if ("link".equals(localName)) {
```
kkrugler: This looks like a separate fix for some issue (vs. what's in the CHANGES.txt entry), yes? What was the `guid` check for previously?

src/main/java/crawlercommons/sitemaps/sax/RSSHandler.java
```diff
@@ -109,47 +107,55 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+            // - a valid absolute URL (not a URN, UUID or similar)
+            // - and no <link> found yet
+            if (locURL == null) {
+                setLocURL();
+            }
+            resetCharacterBuffer();
+        } else if ("item".equals(localName)) {
             maybeAddSiteMapUrl();
-        } else if ("rss".equals(currentElement())) {
+        } else if ("rss".equals(localName)) {
```
kkrugler: Same question about switch from `currentElement` to `localname`.

src/main/java/crawlercommons/sitemaps/sax/RSSHandler.java
```diff
@@ -109,47 +107,55 @@ public void startElement(String uri, String localName, String qName, Attributes
...
             sitemap.setProcessed(true);
+        } else if ("pubDate".equals(localName)) {
+            String value = getAndResetCharacterBuffer();
+            if (value != null) {
```
kkrugler: When would `value` ever be null? And is support for `"pubDate"` something new?

src/main/java/crawlercommons/sitemaps/sax/RSSHandler.java
kkrugler: OK, wait - I think I see what's changing in the logic (`guid` check was in `characters()`).

src/main/java/crawlercommons/sitemaps/sax/DelegatorHandler.java
sebastian-nagel: Ok, I'll update the PR.

src/main/java/crawlercommons/sitemaps/sax/RSSHandler.java
sebastian-nagel: Yes, `guid` was before completely handled in `characters()`, now the content is aggregated in `characters()` and is consumed in `endElement()`.

src/main/java/crawlercommons/sitemaps/sax/RSSHandler.java
sebastian-nagel: > When would `value` ever be null?

Correct, will change it.

> And is support for `"pubDate"` something new?

The code has been moved from `characters()` to `endElement()`.

src/main/java/crawlercommons/sitemaps/sax/RSSHandler.java
sebastian-nagel: Calling `currentElement()` returns the content of `localName` because DelegatorHandler pushes `localName` to the element stack before calling the delegate handler. Only inside `characters()` there is no `localName`. I'll make sure that only `localName` is used in all start/endElement methods, if a value is passed as parameter it should be used instead of accessing the element stack.

sebastian-nagel:
Thanks, @kkrugler!

Issue #81, opened 2015-06-06
The implementation ignores URLs on different subdomains due to urlIsLegal implementation.


suhdev:
The current implementation assumes that a URL is legal only if the sitemap URL is a substring of the URL. This doesn't hold for some websites such as nytimes.com in which the sitemaps are actually on a different subdomain and the urls within those sitemaps points to pages on the main domain. 

```
Snippet from nytimes.com robots.txt: 

Sitemap: http://spiderbites.nytimes.com/sitemaps/www.nytimes.com/sitemap.xml.gz
Sitemap: http://www.nytimes.com/sitemaps/sitemap_news/sitemap.xml.gz
Sitemap: http://spiderbites.nytimes.com/sitemaps/sitemap_video/sitemap.xml.gz

```

Perhaps using the domain name, not the whole sitemap URL can fix this, Google Guava `InternetDomainName` which could help in extracting domain names, or maybe using simple Regex!.

kkrugler:
From http://www.sitemaps.org/protocol.html#location:

> The location of a Sitemap file determines the set of URLs that can be included in that Sitemap. A Sitemap file located at http://example.com/catalog/sitemap.xml can include any URLs starting with http://example.com/catalog/ but can not include URLs starting with http://example.com/images/.

and then:

>  Note that this means that all URLs listed in the Sitemap must use the same protocol (http, in this example) and reside on the same host as the Sitemap. For instance, if the Sitemap is located at http://www.example.com/sitemap.xml, it can't include URLs from http://subdomain.example.com.

So it seems like what you're asking for is support for invalid/broken sitemaps, yes?

Dealing with invalid data (robots.txt, etc) is definitely something we do a lot of, but I think this would want to be handled via a "lenient" vs. "strict" setting.

Chaiavi:
Actually, nytimes are ok, thier sitemap isn't broken.

You can have in your sitemap file, "cross-domain" urls ONLY IF you
authorize those domains in your robots.txt file.


This specific case is mentioned in the same link under the Sitemaps & Cross
Submits subtitle.




This is a feature we don't support yet (although we should consider it
IMHO), it involves parsing the robots.txt file when parsing a sitemap.

kkrugler:
Hi Avi - thanks for the ref...cross submits were added after the original sitemap spec, so I didn't know about them.

I'm still a bit confused about the nytimes.com case. Based on this:

[...]

PR #241, opened 2019-03-26
Robots parser to always handle absolute sitemap URL even without valid base URL

### LINKED ISSUE #240: Robots parser to always handle absolute sitemap URL even without valid base URL
I get an `Invalid URL with sitemap directive` for a sitemap url which is not relative.

I noticed in the implementation [here](https://github.com/crawler-commons/crawler-commons/blob/master/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java#L740) that it only handles relative URLs, which might not be the case always, as now it is preferred to have the absolute URL for the sitemap in `robots.txt`.

Therefore if there is a URL provided with the sitemap, we cannot access it.

sebastian-nagel:
fixes #240

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,7 @@
 Crawler-Commons Change Log
 
 Current Development 1.1-SNAPSHOT (yyyy-mm-dd)
+- [Robots] Robots parser to always handle absolute sitemap URL even without valid base URL (pr3mar, kkrugler, sebastian-nagel) #240
 
 Release 1.0 (2019-03-19)
 - [Sitemaps] Unit tests depend on system timezone (kkrugler, sebastian-nagel) #238
--- a/src/main/java/crawlercommons/robots/BaseRobotsParser.java
+++ b/src/main/java/crawlercommons/robots/BaseRobotsParser.java
@@ -38,7 +38,10 @@ public abstract class BaseRobotsParser implements Serializable {
      * lower-casing, and the prefix match rule.
      * 
      * @param url
-     *            URL that content was fetched from (for reporting purposes)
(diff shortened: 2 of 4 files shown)


kkrugler:
Hi @sebastian-nagel - could you take a pass through https://github.com/crawler-commons/crawler-commons/issues/81 and https://github.com/crawler-commons/crawler-commons/issues/85? I think that in order to correctly implement support for sitemaps in robots.txt, we should complain if the sitemap URL isn't valid, unless we're running in some new lenient mode for robots.txt parsing.

sebastian-nagel:
Hi @kkrugler, aren't the cross-submit issues (#81 and #85) only relevant for checking the sitemap content, i.e., the URLs listed therein? E.g., if we would provide a method `robotsRules.getUrlsFromSitemaps()` which (recursively) fetches and parses the sitemaps, extracts the listed URLs and skips those which are not valid and are also not valid cross-submits. The URL given by `sitemap:` in a robots.txt file cannot be invalid, cf. [Google's description of sitemap: in robots.txt](https://developers.google.com/search/reference/robots_txt#sitemap). Even if spammy sitemaps are listed in the robots.txt ([one example](https://index.commoncrawl.org/CC-MAIN-2018-17/20180426070957/http://livingmoving.com/robots.txt)) - we can only check for cross-submit violations if we process the sitemap(s) and (unfortunately) not when processing the robots.txt.

kkrugler:
Hi @sebastian-nagel - sorry, you're correct, we only return the sitemap URL without worrying about contents of the sitemap itself, so my bad. I shouldn't try to sneak in code reviews while on vacation :)

Issue #255, opened 2019-07-22
robots rules unit tests should be upgraded using junit5 parameterizedtest


Chaiavi:
SimpleRobotRulesParserTest can be upgraded to be smaller, more readable and most important of all - to have all tests separated so instead of using junit4 regular @Test which is used to test sometimes more than 20 urls in one test, we should use junit5 @parameterizedtest thus having each url tested as a test case by itself.

The above gains us the ability to have better granularity of the tests so if a test method fails the other won't topple after it but will be tested separately and should pass.

Chaiavi:
Submitted in 1c74fc6ca057b02401d41319ba10f90df9c96f0b

PR #256, opened 2019-07-22
The robots.txt test now supports the junit5 parameterizedTest, gainin…

Chaiavi:
…g us the ability to have each robot rule be tested separately, so if one rule fails, the others can still show success in their own unit tests.

IMHO it also makes the tests more readable but that is in the eyes of the beholder

This files is smaller by 40 lines, and has 172 unit tests instead of 46 tests


Chaiavi:
This fixes #255

Issue #260, opened 2019-08-19
Upgrade GitIgnore


Chaiavi:
It will be better to upgrade our gitignore file with ignoring all files which should be ignored but created by Java (class files etc), Intellij Idea, Eclipse and maven.

In order to generate the complete list I want to use [the gitignore.io online service](https://www.gitignore.io/)

I intend to append the 4 gitignore files.

Chaiavi:
Closed by #265

Chaiavi:
And by e55bf1f6971bddd142316eb4e009c728541495f2

PR #265, opened 2019-09-04
Updated the gitignore file to include all files which should be ignor…

Chaiavi:
…ed due to the following frameworks: Java, Eclipse, Intellij Idea, Maven.

File generated initially using: https://www.gitignore.io/
This submit attempts to solve: #260

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,188 @@
-.DS_Store
-.classpath
-.project
+# Ignoring files generated from: Java, Eclipse, Intellij Idea, Maven.
+# File generated initially using: https://www.gitignore.io/
+### Eclipse ###
+.metadata
+bin/
+tmp/
+*.tmp
+*.bak
+*.swp
+*~.nib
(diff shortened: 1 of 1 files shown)


Chaiavi:
@jnioche You are right, .DS_Store should remain.
I will fix it.


The other languages you mention, are cases where people have an IDE (like eclipse) which they use to code java or other languages (using the same IDE), thus they have those files generated for them.

But I do get your point, and probably people will use different IDEs for different languages.

But (unless you differ) I want to retain those files which are generated by different plugins although most people don't have those plugins.


By the way, do you understand why the travis build failed? (Regular maven build works on my local machine)

Chaiavi:
ok

- Updated according to the code review of @jnioche 

- Now Travis passes thanks to 559d45cbba6dacad69b5b4eb97f7b000b682d38a

Chaiavi:
This submit closes #260

Issue #264, opened 2019-08-30
Upgrade the toString() method of the BaseRobotRules


Chaiavi:
Currently there is a nice toString in [BaseRobotsRule](https://github.com/crawler-commons/crawler-commons/blob/1c74fc6ca057b02401d41319ba10f90df9c96f0b/src/main/java/crawlercommons/robots/BaseRobotRules.java#L118) method which shows the robots file nicely.

But the toString shows only the first sitemap URL instead of showing all of them.
I understand that this is a calculated decision and not a bug, but as there should not be too many sitemaps in the robots file and if a site has many sitemap it usually will use a sitemap index rather than put all of them in the robots file, so I don't think it will cause any harm if all sitemaps will show in the robots toString.

So if I have a go from the team I will be happy to submit a pull request upgrading the robots toString.

Chaiavi:
Currently the toString looks like this:

> class crawlercommons.robots.SimpleRobotRules:
> no crawl delay
> number of sitemap URLs: 2
>  sitemaps[0]: http://qoutes.over-blog.org/sitemap-news.xml
> number of rules: 4
>Disallow: /mail/subscribe
>Disallow: /mail/valid-*
>Disallow: /contact
>Allow: /

sebastian-nagel:
There should be some limit on the number of sitemaps logged. I've seen robots.txt files ([one example](https://index.commoncrawl.org/CC-MAIN-2018-17/20180426070957/http://livingmoving.com/robots.txt)) listing ten thousands of sitemaps. Calling toString() without any limit eg. for logging may have bad effects. Of course, it's ok to log 10-20 sitemaps and than add a message "remaining N sitemaps skipped" or similar.

Chaiavi:
agreed

PR #268, opened 2019-09-11
Better robots to string

Chaiavi:
Updated the toString method to include a list of sitemaps if there are less than 10 sitemaps in the robots file

This pull request attemps to solve #264

### DIFF
--- a/src/main/java/crawlercommons/robots/BaseRobotRules.java
+++ b/src/main/java/crawlercommons/robots/BaseRobotRules.java
@@ -24,7 +24,6 @@ import java.util.List;
  * Result from parsing a single robots.txt file - which means we get a set of
  * rules, and a crawl-delay.
  */
-
 @SuppressWarnings("serial")
 public abstract class BaseRobotRules implements Serializable {
 
@@ -68,11 +67,6 @@ public abstract class BaseRobotRules implements Serializable {
         return _sitemaps;
     }
 
-    /*
-     * (non-Javadoc)
-     * 
-     * @see java.lang.Object#hashCode()
-     */
     @Override
     public int hashCode() {
         final int prime = 31;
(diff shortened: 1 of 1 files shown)


src/main/java/crawlercommons/robots/BaseRobotRules.java
```diff
@@ -124,14 +112,17 @@ public String toString() {
...
+
         int nSitemaps = getSitemaps().size();
         if (nSitemaps == 0) {
             sb.append(" - no sitemap URLs\n");
         } else {
             sb.append(" - number of sitemap URLs: ").append(nSitemaps).append('\n');
-            sb.append("   sitemaps[0]: ").append(getSitemaps().get(0)).append('\n');
+            if (nSitemaps <= 10) {
+                sb.append(String.join("\n", getSitemaps())).append("\n\n");
+            }
```
kkrugler: Why not append up to 10 sitemaps, versus appending nothing if there or more than 10?

Chaiavi: I followed the same logic which I found in the [SimpleRobotRules](https://github.com/crawler-commons/crawler-commons/blob/e55bf1f6971bddd142316eb4e009c728541495f2/src/main/java/crawlercommons/robots/SimpleRobotRules.java#L369)

But if you think it makes more sense, I can change both files to add the first 10 sitemaps/robotRules

Chaiavi:
I will update this pull request to show 10 sitemaps/robot rules in the case there are more than 10.
As per @kkrugler suggestion

sebastian-nagel:
@Chaiavi, any updates? Need also to rebase and fix the conflicts CHANGES.txt

Chaiavi:
Thanks for the reminder,

I will take the time to finish the task

Chaiavi:
I updated the simple and base robot rules toString() method.

Maven tests passed, but I am working on a new environment and I don't have the use case to test and actually see the results.

Can anybody please test a robots rule with more than 10 sitemaps referenced in it and see the toString method in action ?

That is the problem with abandoning a commit for so long, I lose the context (sorry  :-(  )

PR #272, opened 2019-10-17
[BasicNormalizer] Trailing question mark in url query test should be added

### LINKED ISSUE #247: [BasicNormalizer] Trailing question mark in url query test should be added
Currently our normalizer doesn't remove a trailing question mark - I think it should. [As @sebastian-nagel writes in this thread, the trailing question mark is the expected behaviour]

Example: http://shekhargulati.com?

The above example was taken from [UrlCleaner](https://github.com/shekhargulati/urlcleaner/blob/master/src/test/java/com/shekhargulati/urlcleaner/UrlCleanerSpec.java), test name: shouldRemoveQuestionMarkWhenThereIsNoQueryParameter()

sebastian-nagel:
(fixes #247)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,7 @@
 Crawler-Commons Change Log
 
 Current Development 1.1-SNAPSHOT (yyyy-mm-dd)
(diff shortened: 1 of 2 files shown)


Chaiavi:
Nicely done

The test addition is so small using junit5, it is a pleasure

PR #274, opened 2019-11-09
Make mvn run parameterized tests, fixes #273

sebastian-nagel:
- make test classes public
- upgrade Maven surefire plugin to recent version

PR #275, opened 2019-11-09
[BasicNormalizer] Empty path before query to be normalized to `/`

### LINKED ISSUE #247: [BasicNormalizer] Trailing question mark in url query test should be added

sebastian-nagel:
fixes #247

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,7 +1,7 @@
 Crawler-Commons Change Log
 
 Current Development 1.1-SNAPSHOT (yyyy-mm-dd)
-- [BasicNormalizer] Trailing question mark in url query test should be added (Chaiavi, sebastian-nagel) #247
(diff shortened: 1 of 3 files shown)


kkrugler:
Do we need a change to CHANGES?

Chaiavi:
lgtm too

@sebastian-nagel I think you can merge this one

Maybe update CHANGES in the second pull request ? (#274 ) instead of updating it twice

sebastian-nagel:
I've updated the change log and fixed the title of #247. Thanks for the reviews!

PR #279, opened 2019-12-03
SimpleRobotRulesParser: counter _numWarnings not thread-safe, fixes #278

### LINKED ISSUE #278: SimpleRobotRulesParser: counter _numWarnings not thread-safe
[SimpleRobotRulesParser](http://crawler-commons.github.io/crawler-commons/1.0/crawlercommons/robots/SimpleRobotRulesParser.html) allows to limit the maximum number of warnings logged per parsed robots.txt file. This is useful, otherwise the warnings may flood the log files, esp. in case HTML is returned for the `/robots.txt`.

Warnings are counted in the member field [_numWarnings](https://github.com/crawler-commons/crawler-commons/blob/crawler-commons-1.0/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java#L354). This may cause unexpected behavior in case a SimpleRobotRulesParser instance is used in multiple parallel threads - the crawler implementations I'm familiar with (Nutch and Stormcrawler) use a shared instance of the robots parser. This issue has been observed while analyzing Nutch fetcher log files with tasks running 100+ parallel threads: the logs contain warnings without the corresponding first line mentioning the robots.txt URL.

sebastian-nagel:
- move the counter `_numWarnings` into the ParseState object local to `parseContent(...)`
- after parsing the warnings count is assigned to a member variable of SimpleRobotRulesParser, the value is returned by `getNumWarnings()`. This avoids breaking the API, also the unit tests rely on this method.
- use parameterized method calls for logging, avoid concatenating log messages beforehand

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,7 @@
 Crawler-Commons Change Log
 
 Current Development 1.1-SNAPSHOT (yyyy-mm-dd)
+- SimpleRobotRulesParser: counter _numWarnings not thread-safe (sebastian-nagel, kkrugler) #278
 - ParameterizedTest not executed by mvn builds (sebastian-nagel) #273
 - [BasicNormalizer] Empty path before query to be normalized to `/` (Chaiavi, sebastian-nagel) #247
 - EffectiveTldFinder to validate returned domain names for length restrictions (sebastian-nagel, Chaiavi) #251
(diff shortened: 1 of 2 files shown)


sebastian-nagel:
Hi @kkrugler, my experience with ThreadLocals was that they're significantly slower, every increment requires get the variable from the pool, create a new Integer with the incremented value, and put that back to the pool. Moving the counter to the parseState which is passed down to all functions looked like the better solution.

But we could use ThreadLocal to ensure that the getNumWarnings() returns the correct value. It would be safe, except an executor is used to do the processing and getNumWarnings() is called later outside the executor. I'll update the PR.

sebastian-nagel:
Thanks, @kkrugler. I'll squash the commits, update the changes.txt and merge this PR.

PR #288, opened 2020-06-04
Adding asMap to ExtensionMetadata Interface

### LINKED ISSUE #287: ExtensionMetadata should be able to express itself as a Map
ExtensionMetadata objects should be expressable as a Map<String, String[]>.  This has the benefit of making the data a little more portable for uses in other applications, like StormCrawler.

The originating request comes from [StormCrawler/issue/749](https://github.com/DigitalPebble/storm-crawler/issues/749#issuecomment-631278036) with an implementation recommendation from @sebastian-nagel .

Some implementation details.

- Create an abstract method in `ExtensionMetadata` called `public Map<String, String[]> asMap()`
- Implement the method in classes extending `ExtensionMetadata`, return a `Map<String, String[]>` with all non-null attributes included.

evanhalley:
Adding asMap to ExtensionMetadata interface.  Implemented asMap in all subclasses.

Fixes #287

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/extension/ImageAttributes.java
+++ b/src/main/java/crawlercommons/sitemaps/extension/ImageAttributes.java
@@ -17,7 +17,7 @@
 package crawlercommons.sitemaps.extension;
 
 import java.net.URL;
-import java.util.Objects;
+import java.util.*;
 
 /**
  * Data model for Google extension to the sitemap protocol regarding images
@@ -128,4 +128,29 @@ public class ImageAttributes extends ExtensionMetadata {
                         && Objects.equals(license, that.license);
     }
 
+    @Override
+    public Map<String, String[]> asMap() {
+        Map<String, String[]> map = new HashMap<>();
+
+        if (loc != null) {
+            map.put("loc", new String[]{ loc.toString() });
+        }
+
+        if (caption != null) {
+            map.put("caption", new String[]{ caption });
+        }
+
+        if (geoLocation != null) {
+            map.put("geo_location", new String[]{ geoLocation });
+        }
+
+        if (title != null) {
+            map.put("title", new String[]{ title });
+        }
+
+        if (license != null) {
+            map.put("license", new String[]{ license.toString() });
+        }
+        return Collections.unmodifiableMap(map);
+    }
 }
--- a/src/main/java/crawlercommons/sitemaps/extension/ImageAttributes.java
+++ b/src/main/java/crawlercommons/sitemaps/extension/ImageAttributes.java
@@ -17,13 +17,23 @@
 package crawlercommons.sitemaps.extension;
 
 import java.net.URL;
-import java.util.*;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
 
 /**
  * Data model for Google extension to the sitemap protocol regarding images
  * indexing, as per http://www.google.com/schemas/sitemap-image/1.1
  */
 public class ImageAttributes extends ExtensionMetadata {
+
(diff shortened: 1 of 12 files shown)


src/main/java/crawlercommons/sitemaps/extension/ImageAttributes.java
```diff
@@ -128,4 +128,29 @@ public boolean equals(Object other) {
                         && Objects.equals(license, that.license);
     }
 
+    @Override
+    public Map<String, String[]> asMap() {
+        Map<String, String[]> map = new HashMap<>();
+
+        if (loc != null) {
+            map.put("loc", new String[]{ loc.toString() });
```
kkrugler, line 136: For this, and all other xxxAttributes classes, are there well defined map key name constants we could use? 

If not, then I would be in favor of adding them as public constants to the xxxAttribute classes, to avoid raw string constant usage in consumer code.

src/main/java/crawlercommons/sitemaps/extension/LinkAttributes.java
```diff
@@ -96,4 +98,20 @@ public boolean equals(Object other) {
...
+        Map<String, String[]> map = new HashMap<>();
+
+        if (href != null) {
+            map.put("href", new String[]{ href.toString() });
+        }
+
+        if (params != null) {
+
+            for (Entry<String, String> entry : params.entrySet()) {
+                map.put(entry.getKey(), new String[] { entry.getValue() });
```
kkrugler, line 112: Are we guaranteed that non of these entries has a key = `href`? I think it would be safer and clearer to make the map key something like `params.xxx`

src/main/java/crawlercommons/sitemaps/extension/ImageAttributes.java
evanhalley: There are not constants that I could fine.  I'll add some and use those instead of strings.

src/main/java/crawlercommons/sitemaps/extension/LinkAttributes.java
evanhalley: Good call out, it is not a guarantee.  I'll prefix these parameters to avoid conflicts and potential overwrites.

evanhalley:
@kkrugler This should be ready for another review.  I addressed both points of feedback.  Thanks again.

src/main/java/crawlercommons/sitemaps/extension/LinkAttributes.java
```diff
@@ -32,6 +34,10 @@
  * you to check for mistakes.</blockquote>
  */
 public class LinkAttributes extends ExtensionMetadata {
+
+    public static final String HREF = "href";
+    public static final String PARAMS = "params.%s";
```
kkrugler, line 39: My only comment here is that `PARAMS` is a bit different from the other keys. Not a big deal, but I'd probably call it `PARAMS_PREFIX`, and make its value just `params.`

In any case I'd add a comment explaining how it's used for the map key.

evanhalley: Thinking about PARAMS_PREFIX a little more, does it make sense to decrease the visiblity?  I'm not sure how useful it is outside of the class.

jnioche:
hi @evanhalley, is more work needed on this or is it ready to be merged? It now passes the tests, thanks

evanhalley:
@jnioche this is ready to be merged.

jnioche:
thanks again @evanhalley, your contribution is greatly appreciated

evanhalley:
@jnioche @kkrugler thanks for the help!

PR #293, opened 2020-06-19
[sitemaps] Allow to deduplicate sitemap links in sitemap indexes, fixes #262

### LINKED ISSUE #262: [sitemaps] Allow to deduplicate sitemap links in sitemap indexes
(see [NUTCH-2730](https://issues.apache.org/jira/browse/NUTCH-2730))

A sitemap index may list the same sitemap multiple times (erroneously or maliciously). There should be an easy way to deduplicate the list of subsitemaps
- provide a method to that (eg. `getSitemaps(boolean unify)`)
- or overwrite the "equals" and "hashCode" methods in AbstractSitemap to make the deduplication easy

sebastian-nagel:


jnioche:
thanks @sebastian-nagel

Chaiavi:
Why would anyone want a duplicate url in his sitemaps?
Is this really a feature we need ?

sebastian-nagel:
Just as a matter of fact: there are sitemap indexes out there in the wild which contain the same URL pointing to a sitemap multiple times. I doubt, that anybody "wants" this, it could be a bug in the sitemap generator. And a crawler then should smartly ignore these erroneous duplicates. This method helps a little bit to implement a cautious sitemap processor, at least, it eventually makes crawler developers aware of a potential problem.

Chaiavi:
I see, you are right.

it is not our place to fix their buggy generator, we should give them a
sitemap object representation of their actual sitemap.


Thanks for the explanation.

Issue #300, opened 2020-07-16
For VideoAttributes.asMap(), Duration not included in map


evanhalley:
When VideoAttributes.asMap is called, the duration is not added to the map (oversight on my part during implementation of #287.  

The duration should be added.

PR #301, opened 2020-07-16
Adding duration to the map generated by VideoAttributes.asMap

evanhalley:
Adding duration to the map generated by VideoAttributes.asMap.  This is a fix for #300


jnioche:
thanks @evanhalley

PR #302, opened 2020-08-07
[EffectiveTldFinder] Download latest effective_tld_names.dat during Maven build

### LINKED ISSUE #295: EffectiveTldFinder: upgrade public suffix list
The public suffix list packed with crawler-commons 1.1 dates to March 2019, it should be upgraded, ideally shortly before the release of 1.2. See #219 (40531ef).

rzo1:
- introduces download-maven-plugin to download the `effective_tld_names.dat `during the maven build
- ensures no cached effective_tld_names.dat is used for the `release` profile

Contributes / fixes #295


sebastian-nagel:
+1 Looks good:
 - the public suffix list is cached after the first download by the maven download plugin (otherwise a build after a clean would fail if there is not internet connection)
 - good idea to ignore the cache when preparing a release

Important to know because of the frequent changes of the [public suffix list](/publicsuffix/list):
- there is a small chance that unit tests fail after an update of the list
- if a 100% stable mapping of host to domain names is required, developers should use their own copy of the list and explicitly initialize EffectiveTldFinder using the custom list

But I assume that most users want an up-to-date list - what is ensured by this PR. Thanks!

PR #306, opened 2020-11-11
Normalize URL without a scheme, resolves #271

### LINKED ISSUE #271: Normalize URL without a scheme
I mean that if someone attempts to normalize "google.com" should our normalizer return null (like it does today) or should it should we return http://google.com

Maybe in the aggresive normalizer ?

sebastian-nagel:
- prefix URL without a scheme with `http://`

### DIFF
--- a/src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
+++ b/src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
@@ -60,6 +60,12 @@ public class BasicURLNormalizer extends URLFilter {
      */
     private final static Pattern unescapeRulePattern = Pattern.compile("%([0-9A-Fa-f]{2})");
 
+    /**
+     * Match URLs starting with a valid scheme, see
+     * https://tools.ietf.org/html/rfc2396#section-3.1
+     */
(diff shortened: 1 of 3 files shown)


src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
```diff
@@ -143,6 +149,15 @@ public String filter(String urlString) {
...
+            if (!hasSchemePattern.matcher(urlString).find()) {
+                // no protocol/scheme : try to prefix http://
+                try {
+                    url = new URL("http://" + urlString);
+                } catch (MalformedURLException e1) {
+                }
+            }
+        }
+        if (url == null) {
             LOG.info("Malformed URL {}", urlString);
```
kkrugler, line 161: Not related to your code change, but logging at info level here seems wrong - I would change to debug

kkrugler (approved):
lgtm - one comment about logging level (for unchanged code)

Chaiavi (approved):
Perfect, thank you S

src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
sebastian-nagel: Ok, done.

PR #307, opened 2020-12-08
Allow to normalize URLs in sitemaps, resolves #305

### LINKED ISSUE #305: [sitemaps] Allow to normalize URLs in sitemaps
Url strings  without protocol will produce MalformedURLException. Add protocol to url string if it missing and not strict mode.
Example: http://www.clasohlson.com/sitemap.xml

```
private void maybeAddSiteMap() {
        if (loc == null) {
            return;
        }
        String locUrlString = stripAllBlank(loc);
        //url string without protocol will produce MalformedURLException 
        if (!isStrict()) { //maybe add new flag that user can specify if we should auto assign protocol 
            if (!locUrlString.isEmpty() && !locUrlString.contains("://")) { // simple check if we have "://" in string, or maybe regex that handles ALL protocols what is hard [http, https, ftp, ssl...]
                locUrlString = sitemap.getUrl().getProtocol() + "://" + locUrlString; // get protocol from the sitemap
            }
        }
        try {
            // check that the value is a valid URL
            URL locURL = new URL(locUrlString);
            SiteMap s = new SiteMap(locURL, lastMod);
            sitemap.addSitemap(s);
            LOG.debug("  {}. {}", (i + 1), s);
        } catch (MalformedURLException e) {
            LOG.trace("Don't create an entry with a bad URL", e);
            LOG.debug("Bad url: [{}]", loc);
        }
        loc = null;
        locClosed = false;
        lastMod = null;
    }
```

```
private void maybeAddSiteMapUrl() {
        if (valid) {
            if (loc == null) {
                LOG.debug("Missing url");
                LOG.trace("Can't create an entry with a missing URL");
            } else {
                //no need to check if we should assign protocol, already checked with "valid" flag
                String locUrlString = stripAllBlank(loc.toString());
                if (!locUrlString.isEmpty() && !locUrlString.contains("://")) {
                    locUrlString = sitemap.getUrl().getProtocol() + "://" + locUrlString;
                }
                SiteMapURL sUrl = new SiteMapURL(locUrlString, lastMod, null, null, valid);
                sitemap.addSiteMapUrl(sUrl);
                LOG.debug("  {}. {}", (++i), sUrl);
            }
        }
        loc = null;
        lastMod = null;
    }
```

Duplicate: https://github.com/crawler-commons/crawler-commons/issues/271

sebastian-nagel:
- extend SiteMapParser by methods to register a URLFilter (function) used to normalize or filter (if null is returned) URLs found in   sitemaps
- implement URL filtering in sitemap parsers / XML handlers
- add unit tests to verify URL filtering for text and XML sitemaps

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
+++ b/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
@@ -34,6 +34,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.zip.GZIPInputStream;
 
 import javax.xml.parsers.ParserConfigurationException;
@@ -48,6 +49,7 @@ import org.xml.sax.EntityResolver;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 
+import crawlercommons.filters.URLFilter;
 import crawlercommons.mimetypes.MimeTypeDetector;
 import crawlercommons.sitemaps.AbstractSiteMap.SitemapType;
 import crawlercommons.sitemaps.extension.Extension;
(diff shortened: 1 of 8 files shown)


Chaiavi (approved):
lgtm

Issue #308, opened 2021-01-04
[BasicNormalizer] Query parameters normalization in BasicURLNormalizer


aecio:
I've been working on a PR for #246 and found the following normalization test cases in [weirdToNormalizedUrls.csv](https://github.com/crawler-commons/crawler-commons/blob/master/src/test/resources/normalizer/weirdToNormalizedUrls.csv) that I believe are problematic:
- `http://example.com?` => `http://example.com/?`
- `http://foo.com/aa?referer=http://bar.com` =>`http://foo.com/aa?referer=http://bar.com`

Shouldn't these be normalized as follows?
- `http://example.com?` => `http://example.com/`
- `http://foo.com/aa?referer=http://bar.com` =>`http://foo.com/aa?referer=http%3A%2F%2Fbar.com`

In addition, well-formed URLs with empty query parameters could have the question mark removed as follows:
- `http://example.com/?` => `http://example.com/`

Any thoughts about this? I'll submit a pull request that will include a fix for these cases.

PR #309, opened 2021-01-04
Query parameters normalization in BasicURLNormalizer

### LINKED ISSUE #246: [BasicNormalizer] sorting the Query Parameters
When normalizing a URL should these two urls be normalized to the same url ?
http://shekhargulati.com?lang=en&article=fred
http://shekhargulati.com?article=fred&lang=en

This can be achieved by sorting the query params.

The above example was taken from [UrlCleaner](https://github.com/shekhargulati/urlcleaner/blob/master/src/test/java/com/shekhargulati/urlcleaner/UrlCleanerSpec.java)

It is an actual unit test there (one which we fail :-(   ), the test name is: shouldSortQueryParameters

aecio:
- Sort query parameters (fix #246)
- Allows to (optionally) remove common irrelevant query parameters

Related to #308.


src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
```diff
@@ -255,6 +262,162 @@ public String filter(String urlString) {
         return urlString;
     }
 
+    private static URL parseUrl(String urlString) {
```
Chaiavi, line 265: I would really appreciate here a JavaDoc explaining what this method is doing (Even though it is private)

src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
```diff
@@ -255,6 +262,162 @@ public String filter(String urlString) {
...
+                try {
+                    url = new URL("http://" + urlString);
+                } catch (MalformedURLException e1) {
+                }
+            }
+        }
+        return url;
+    }
+
+    private String normalizeUrlFile(String file) {
```
Chaiavi, line 281: Javadoc?

src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
```diff
@@ -255,6 +262,162 @@ public String filter(String urlString) {
...
+            pairs.sort(NameValuePair.NAME_COMPARATOR);
+            normalizedFile
+                .append('?')
+                .append(formatQueryParameters(pairs));
+        }
+
+        return normalizedFile.toString();
+    }
+
+    public static List<NameValuePair> parseQueryParameters(final String s, final int queryStartIdx,
```
Chaiavi, line 315: JavaDoc

src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
```diff
@@ -255,6 +262,162 @@ public String filter(String urlString) {
...
+
+    private static boolean isValueEnd(final char c) {
+        return c == '&' || c == ';';
+    }
+
+    private static boolean isNameEnd(final char c) {
+        return c == '=' || c == '&' || c == ';';
+    }
+
+    public static String formatQueryParameters(final List<NameValuePair> parameters) {
```
Chaiavi, line 382: JavaDoc

Chaiavi (commented):
Format / Parse are good for method names but are not descriptive enough so I don't understand what the method does unless I dive into the code (versus "isBlank" which is ok without javadoc)

src/main/java/crawlercommons/utils/Strings.java
```diff
@@ -0,0 +1,28 @@
+package crawlercommons.utils;
+
+/**
+ * Util functions for manipulating strings.
+ */
+public class Strings {
+
+    public static boolean isBlank(final String cs) {
+        if (cs == null || cs.isEmpty()) {
```
Chaiavi, line 9: Change cs.isEmpty()
to
cs.trim().isEmpty()

And you will be able to cut several lines from the code

src/main/java/crawlercommons/utils/Strings.java
```diff
@@ -0,0 +1,28 @@
...
+        }
+        for (int i = 0; i < cs.length(); i++) {
+            if (Character.isWhitespace(cs.charAt(i)) == false) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    public static boolean isNotBlank(final String cs) {
```
Chaiavi, line 20: !sure this one is needed
it is clear enough to use !isBlank(...) in the code, no ?

src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
aecio: done.

src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
aecio: done.

src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
aecio: done.

src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
```diff
@@ -305,7 +519,7 @@ private String unescapePath(String path) {
         // Traverse over all encoded groups
         while (matcher.find()) {
             // Append everything up to this group
-            sb.append(path.substring(end + 1, matcher.start()));
+            sb.append(path, end + 1, matcher.start());
```
aecio, line 522: In case anyone is wondering about this change, these lines are equivalent. Intellij automatically suggests this refactoring, which avoids one extra function call.

src/main/java/crawlercommons/utils/Strings.java
aecio: Can you clarify what lines exactly could be cut? The method `trim()` might unnecessarily allocate and copy bytes into a new String object. Also, I'm not sure if `trim()` removes the same chars as `Character.isWhitespace()`. This implementation is similar to the StringUtils class used in apache `http-components` which used in their URL parsing code.

aecio (commented):
Thanks for the review, @Chaiavi. I addressed your comments and tried to improved comments and JavaDocs.

src/main/java/crawlercommons/utils/Strings.java
aecio: I don't have strong options here, so removed it.

src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
aecio: done.

src/main/java/crawlercommons/utils/Strings.java
aecio: Can you clarify what lines exactly could be cut? The method `trim()` might unnecessarily allocate and copy bytes into a new String object. Also, I'm not sure if `trim()` removes the same chars as `Character.isWhitespace()`. This implementation is similar to the `StringUtils` class used in Apache's http-components library URL parsing code.

jnioche:
Nice one @aecio 
Could you please add the copyright headers to the new classes + add a line to CHANGES.txt?
Sorry if this sounds a bit formal. I haven't had time to review in details yet

sebastian-nagel (commented):
Thanks, @aecio. The PR includes a lot of useful normalization steps. I know it isn't easy to balance between strict normalization (eg. to avoid duplicates) and the preservation of potential semantically different URLs.

aecio (commented):
Thanks for the comments and the links, @sebastian-nagel.

src/test/resources/normalizer/weirdToNormalizedUrls.csv
```diff
@@ -124,7 +124,7 @@ http://foo.com/aa/bb//foo.html, http://foo.com/aa/bb/foo.html
 http://foo.com//aa//bb//foo.html, http://foo.com/aa/bb/foo.html
 http://foo.com////aa////bb//foo.html, http://foo.com/aa/bb/foo.html
 http://foo.com////aa////bb////foo.html, http://foo.com/aa/bb/foo.html
-http://foo.com/aa?referer=http://bar.com, http://foo.com/aa?referer=http://bar.com
```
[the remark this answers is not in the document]
aecio: I was working on this change and found the following existing test cases which seem to be inconsistent:

1. `http://x.com/show?http%3A%2F%2Fx.com%2Fb` => `http://x.com/show?http%3A%2F%2Fx.com%2Fb`
2. `http://foo.com/aa?referer=http://bar.com` => `http://foo.com/aa?referer=http://bar.com`

Even though the first case is malformed with an incomplete query name-value pair,  the string `http%3A%2F%2Fx.com%2Fb` should likely be interpreted as a query name. If we are not encoding ':' and '/', then shouldn't the correct normalization be: 
`http://x.com/show?http%3A%2F%2Fx.com%2Fb` => `http://x.com/show?http://x.com/b`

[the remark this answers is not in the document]
sebastian-nagel: Hi @aecio, the query part of a URL is not necessarily filled from a form. This means it can contain any arbitrary query string which isn't a key-value structure and with a encoding not (fully) consistent with `application/x-www-form-urlencoded`. Just a few examples:
- http://etab.ac-poitiers.fr/lepmo/spip.php?article247
- https://norma-jackson-fine-art.com/gallery/index.php?/category/52
- https://cmsimplexh.webdesign-keil.de/?Fachthemen/Ein_CMSimple_XH_Template_erstellen/Idee_%2F_Planung - here `/` and `%2F` have a different semantic replacing them cause a [page not found](https://cmsimplexh.webdesign-keil.de/?Fachthemen%2FEin_CMSimple_XH_Template_erstellen%2FIdee_%2F_Planung)

Of course, most often the [query string](https://en.wikipedia.org/wiki/Query_string) stems from a HTML form. However, it's also important not to break valid URLs. That's what I mean with finding the "balance between strict normalization (eg. to avoid duplicates) and the preservation of potential semantically different URLs".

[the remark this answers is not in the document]
aecio: I understand we should not change the URL when there is potential for breaking it. Your last example shows an interesting edge case where using a standard decoding could create problems. What is still not clear to me is what is the minimal set of normalizations that are safe to apply to the query data. 
- Should we not change the data query data at all?
- When should we sort name-value pairs? When doing so, should we keep data encoding intact?
- Is there any case we want to change the data encoding (maybe only when there are illegal characters)?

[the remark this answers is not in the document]
sebastian-nagel: > the minimal set of normalizations that are safe to apply to the query data
> Is there any case we want to change the data encoding...

See [RFC3986](https://tools.ietf.org/html/rfc3986#section-3.4) and the [whatwg URL spec](https://url.spec.whatwg.org/#query-percent-encode-set) mentioned above. There's an interesting discussion in whatwg/url#369 and a nice JavaScript function to test how browsers normalize URLs.

> Should we not change the data query data at all?

What about making this optional (maybe on by default)? Your PR already adds a constructor to configure the removal of parameters by name. Any good ideas how to make the URL normalization configurable?

Chaiavi:
Lots of work has gone into this

Where do we stand here ?
Can anyone summarize what is still missing (@aecio , @sebastian-nagel) ?

src/test/resources/normalizer/weirdToNormalizedUrls.csv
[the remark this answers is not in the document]
aecio: To make the URL normalization more configurable, we can change the constructor to a builder object that contains all configurable options. This would allow adding many options without having a complex constructor and while still keeping a simple fluent API with reasonable defaults options.

I think that for this case, the options could be the following?

    urlEncodeQueryParameters = NEVER | ALWAYS | KEY_VALUE_PAIRS_ONLY

If so, what would be the desired default? `KEY_VALUE_PAIRS_ONLY`?

src/test/resources/normalizer/weirdToNormalizedUrls.csv
```diff
@@ -178,3 +178,39 @@ https://www.example.org./, https://www.example.org/
...
 file:/var/www/html/foo/../bar/index.html, file:/var/www/html/bar/index.html
+
+# check that parameters are sorted
+http://example.com/?b=1&a=1, http://example.com/?a=1&b=1
+http://foo.com/foo.html?b=1&a=1, http://foo.com/foo.html?a=1&b=1
+http://foo.com/index?a=1&b=2, http://foo.com/index?a=1&b=2
+http://foo.com/index?b=2&a=1, http://foo.com/index?a=1&b=2
+http://foo.com/index?b=2&a=1#c, http://foo.com/index?a=1&b=2
+
+# normalize query parameters separated by semicolons
```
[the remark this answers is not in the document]
aecio: I'd be happy to just remove this, but we could also add it as a configurable option.

aecio:
@Chaiavi As I recall, this PR is doing more normalization than browsers typically do (and than some users would like by default). I was doing this on free time and eventually got too busy, but I'd be happy to continue when I get the chance if folks think this PR is going in the right direction.

sebastian-nagel:
> doing more normalization than browsers typically do

I would also opt to keep the defaults on the safe side and let users actively toggle on the more aggressive options. 

> this PR is going in the right direction

But otherwise yes, it's the right direction.

sebastian-nagel (approved):
+1 Looks good. I'd opt to merge this PR now and opened #321 to address the configuration of the BasicURLNormalizer by providing a builder. Thanks, @aecio!

src/test/resources/normalizer/weirdToNormalizedUrls.csv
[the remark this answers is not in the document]
aecio: This has been removed.

aecio:
Thanks, for reviewing @sebastian-nagel. I still have some uncommitted changes where I was trying to follow the URL specs (RFC3986 and whatwg URL) more closely during URL encoding to avoid unsafe normalizations (which turns out to be a bit tricky). Before merging this: what about I reduce the scope of this PR and split it into two: one with query parameter sorting and removing common irrelevant query parameters, which could be merged now; and another PR with the unsafe URL encoding normalizations, which can be merged later (maybe after the builder is introduced)?

sebastian-nagel:
Yes, I definitely agree: if there is still work to do, better split this PR into smaller work packages. Easier to review, discuss, and less danger that complex PRs lie around until they get finally merged. So, I'll wait for you updates on this PR. Let us know if you want to work on #321, otherwise I'd try to open a PR.

aecio:
Ok, I cleaned up all the extra URL encoding code and use the existing escape/unescape methods instead. Also, added some test cases that cover URLs that have data in the query that should not be encoded. I'll try to open a new PR later only with the URL encoding part.

I'd be happy to work on #321 as well, but won't be able to do it until the weekend or maybe next week. So feel to do it if you'd rather get it done sooner.

sebastian-nagel:
Thanks, @aecio!

Rebased, squashed and fixed the sitemaps unit tests using the URL normalizer and now failing on sorted query params.

PR #313, opened 2021-08-09
Normalizing CHANGES.txt

### LINKED ISSUE #270: Normalize CHANGES.txt
CHANGES.txt currently holds lines using different Prefixes/suffixes etc.

I suggest normalizing this file using a line-template we should agree upon

For example
- [UNIT_NAME] ISSUE_TITLE (SUBMITTER/S) #ISSUE_NUMBER

Chaiavi:
This normalization basically adds the [Unit_Name] in front of the issue when it is obvious and when it is missing

Added the [Domains] unit name (as in the java package name)
Didn't touch the issues changelog prior to v0.7

This resolves #270


Chaiavi:
> Hi @Chaiavi, the formatting looks good.
> 
> Just some thoughts about the unit names:
> 
> * "Normalizer" may raise the question from outsiders which types of items are normalized? Maybe "URLFilters" (the API interface) is preferable? Or even just "URLs"?
> * not sure whether "Robots.txt" would be better than "Robots" alone? Definitely more precise and no risk of confusion - "robot" is somewhat synonymous to "crawler".
> * "Sitemaps" and "Domains" should be definitely clear in the context of web crawling

Thanks @sebastian-nagel,

- I see your point about "Normalizer", although I don't like the URLFilters as it doesn't actually filter URLs, it normalizes them, so I will upgrade to your second suggestion of "URLs"

- About Robots.txt, I will update it and see if it hurts my eyes too much or not  :-)

sebastian-nagel:
+1 looks fine!

sebastian-nagel (approved):

PR #323, opened 2021-09-29
Fix XXE vulnerability in Sitemap parser

kovyrin:
We've been using the Crawler Commons library in our product for a while now and recently received a vulnerability report from a security researcher. Upon investigation, we have discovered that the issue was caused by an insufficiently secure SAX parser configuration within the SitemapParser class.

In this PR I have added a test case to demonstrate an XXE attack by reading a local file via the SitemapParser. I've made a change in the parser to disable external entity resolution according to the [recommendations from OWASP](https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#java) and now the tests work as expected, the invalid URL with an external entity is getting ignored during parsing. I've tried to enable the "http://apache.org/xml/features/disallow-doctype-decl" feature, but then sitemap parsing completely blows up when any DOCTYPE definitions are present and I wasn't sure we want to do that. Let me know if we actually want to go that far and completely disable doctype resolution (which IMO would be reasonable for sitemap parsing, but I'll leave it up to you).

Let me know if I could improve the pull request in any way.

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
+++ b/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
@@ -589,7 +589,11 @@ public class SiteMapParser {
         // support the use of an explicit namespace.
         factory.setNamespaceAware(true);
 
+        // Configure underlying parser features to reduce the risk of XXE attacks
+        // See https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#java
         try {
+            factory.setFeature("http://xml.org/sax/features/external-general-entities", false);
+            factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
             factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
         } catch (Exception e) {
             throw new RuntimeException("Failed to configure XML parser: " + e.toString());
--- a/src/test/java/crawlercommons/sitemaps/SiteMapParserTest.java
+++ b/src/test/java/crawlercommons/sitemaps/SiteMapParserTest.java
@@ -116,10 +116,8 @@ public class SiteMapParserTest {
 
     @Test
     public void testSitemapXXE() throws UnknownFormatException, IOException {
-        // Create a temporary file on disk that would be read if we were vulnerable to XXE
-        Path doNotVisit = Files.createTempFile("do-not-visit", ".txt");
-        doNotVisit.toFile().deleteOnExit();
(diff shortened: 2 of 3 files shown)


sebastian-nagel:
+1 Thanks, @kovyrin! Problem and solution confirmed.

- I'm tempted to copy/modify the unit test to also ensure that the XInclude is always disabled (it already is)

- the unit test uses a temp file - are there any platforms known where this could be an issue? We could use a static file in `src/test/resources/sitemaps/` instead?

> I've tried to enable the "http://apache.org/xml/features/disallow-doctype-decl" feature, but then sitemap parsing completely blows up when any DOCTYPE definitions are present and I wasn't sure we want to do that. Let me know if we actually want to go that far and completely disable doctype resolution (which IMO would be reasonable for sitemap parsing, but I'll leave it up to you).

Good point. For XML sitemaps that's hardly required but we use the XML parser also to parse feeds including RDF and those may include DOCTYPE definitions. Feed formats are allowed as "[other formats](https://www.sitemaps.org/protocol.html#otherformats)" in the sitemap spec.

I'll also run the parser over a large collection of sitemaps to verify that there are no regressions.

sebastian-nagel:
> run the parser over a large collection of sitemaps

I've found no regressions.

kovyrin:
> We could use a static file in src/test/resources/sitemaps/ instead?

I can make that change tomorrow if you think we need it :-)

kovyrin:
Updated the test code to not rely on temporary files

sebastian-nagel:
Thanks, @kovyrin! I'll do some final tests and merge the PR.

sebastian-nagel:
Thanks again, @kovyrin! I've added also a unit test to make sure the parser is not vulnerable to XInclude attacks (copied mostly from your test).

kovyrin:
Do you have any estimate on when we could expect the changes to be released?

sebastian-nagel:
> when we could expect the changes to be released?

I hope very soon, maybe next week. The [release of 1.2 is already discussed](https://groups.google.com/g/crawler-commons/c/bIOUgR0qRaU/m/wKJkR41oAgAJ).

sebastian-nagel:
Hi @kovyrin, the [vote for the 1.2 release](https://groups.google.com/g/crawler-commons/c/A8uKEFWxU3Y/m/H4SO7M4FBgAJ) is open.

kovyrin:
Closing the loop: upgraded the library in our project ([this thing](https://www.elastic.co/guide/en/app-search/current/web-crawler.html)) and all of our tests pass, including the ones we have added specifically for XXE in sitemaps.

Issue #96, opened 2015-09-14
Remove HTTP fetcher support


kkrugler:
Once we've let release 0.6 bake for a while, we should remove the deprecated fetcher support.

jnioche:
The next release will be 0.7 and that's where the fetcher classes will be marked as deprecated. We can remove them altogether in 0.8

jnioche:
@kkrugler do you want to take this one or are you happy for me to do it?

kkrugler:
I'll take care of it, thanks.

PR #342, opened 2021-10-19
Update pom.xml to address Maven warnings and deprecations

sebastian-nagel:
- remove deprecated parent, see https://central.sonatype.org/publish/publish-maven/#deprecated-oss-parent
- update links to XML schema
- remove Jetty dependency not used anymore since Fetcher moved to separate project (cf. #96)
- fix surefire plugin parameters
- fix indentation


Chaiavi (approved):
All seems good

Always good to remove an unused dependency

Thanks S.N.

Issue #245, opened 2019-07-02
Implement Robots Exclusion Protocol (REP) IETF Draft


sebastian-nagel:
See https://tools.ietf.org/html/draft-rep-wg-topic-00

Google ([REP blog post](https://webmasters.googleblog.com/2019/07/rep-id.html)) published an [Apache licensed robots.txt parser](https://github.com/google/robotstxt) written in C++. It ships with tests which we could port first to see how the existing SimpleRobotRulesParser conforms to the REP draft.

sebastian-nagel:
[RFC 9309](https://www.rfc-editor.org/rfc/rfc9309.html) is now a proposed standard.

Added: #389, #390.

sebastian-nagel:
The unit tests of the RFC 9309 reference parser are ported by #360. Several other issues / PRs make SimpleRobotRulesParser comply with RFC 9309. Resolving this issue as "implemented".

Issue #192, opened 2018-01-02
[Robots.txt] Matching user-agent names does not conform to robots.txt RFC


YossiTamari:
http://www.csurams.com/robots.txt contains the lines:
_User-agent: Mozilla/5.0 (compatible; Butterfly/1.0; +http://labs.topsy.com/butterfly/) Gecko/2009032608 Firefox/3.0.8
Disallow: /_

SimpleRobotRulesParser parses this as disallowing access to any agent name that **starts with** “Mozilla/5.0”. Based on the [robots.txt RFC](http://www.robotstxt.org/norobots-rfc.txt), this should not be the case: only if the whole name of the agent is contained in the User-agent line, does the rule apply. The problem stems from the fact that that the User-agent line is split into "words", and this seems to be contrary to the RFC.

sebastian-nagel:
Thanks, @YossiTamari.

The intention was obviously to exclude the [Topsy](https://web.archive.org/web/20120804023136/http://topsy.com/)'s "Butterfly" bot. Of course, just copying the user-agent string from the webserver logs to the robots.txt isn't the right approach and wasn't intention for the "User-agent line name token" in the  [robots.txt RFC](http://www.robotstxt.org/norobots-rfc.txt):

> 3.2.1 The User-agent line
>   Name tokens are used to allow robots to identify themselves via a
>   simple product token. Name tokens should be short and to the
>   point. The name token a robot chooses for itself should be sent
>   as part of the HTTP User-agent header, and must be well documented.

Btw., there is also no reason why you have to use the user-agent name sent in the HTTP header also for robots.txt parsing: in Nutch you would set `http.robots.agents` to `butterfly` and `http.agent.name` to `Mozilla/5.0 ...`. That may solve your problem in the concrete case if you have chosen a unique agent name. You hardly want to name your polite bot "Mozilla".

[...]

PR #351, opened 2021-12-14
Improve robots check draft rfc compliance

yoeduardoj:
This library suffers from multiple issues in processing robots.txt matches when compared to what's considered to be standard behavior. 

https://datatracker.ietf.org/doc/html/draft-rep-wg-topic-00 

- It doesn't merge multiple blocks for the same user agent. 
```
Crawlers MUST find the group that matches the product token exactly, 
   and then obey the rules of the group. If there is more than one 
   group matching the user-agent, the matching groups' rules MUST be 
   combined into one group. 
```
- It doesn't evaluate all matches to find the most specific. 
```
The most specific match found MUST be 
   used. The most specific match is the match that has the most octets. 
```
- When in conflict, it doesn't resolve conflicts correctly (because it doesn't eval them) 
```
If an allow and disallow rule is equivalent, the allow SHOULD be 
   used. 
```

This patch tries to address all of these issues.

### DIFF
--- a/src/main/java/crawlercommons/robots/SimpleRobotRules.java
+++ b/src/main/java/crawlercommons/robots/SimpleRobotRules.java
@@ -163,13 +163,36 @@ public class SimpleRobotRules extends BaseRobotRules {
                 return true;
             }
 
+            boolean isAllowed = true;
+            int longestRuleMatch = Integer.MIN_VALUE;
             for (RobotRule rule : _rules) {
-                if (ruleMatches(pathWithQuery, rule._prefix)) {
-                    return rule._allow;
+                int matchLength = ruleMatches(pathWithQuery, rule._prefix);
+                if (matchLength == -1) {
+                    // See precedence-of-rules test case for an example
+                    // Some webmasters expect behavior close to google's, and this block is equivalent to:
+                    // https://github.com/google/robotstxt/blob/02bc6cdfa32db50d42563180c42aeb47042b4f0c/robots.cc#L605-L618
+                    // There are example robots.txt in the wild that benefit from this.
+                    // As of 2/7/2022, https://venmo.com/robots.txt for instance.
+                    if (rule._prefix.endsWith("index.htm") || rule._prefix.endsWith("index.html")) {
+                        matchLength = ruleMatches(pathWithQuery, rule._prefix.substring(0, rule._prefix.indexOf("index.htm")) + "$");
+                        if (matchLength == -1) {
+                            continue;
+                        }
+                    } else {
+                        continue;
+                    }
                 }
+
+                if (longestRuleMatch < matchLength) {
+                    longestRuleMatch = matchLength;
+                    isAllowed = rule.isAllow();
+                } else if (longestRuleMatch == matchLength) {
+                    isAllowed |= rule.isAllow();
+                }
+                // else we've already got a more specific rule, and this match doesn't matter
             }
 
-            return true;
+            return isAllowed;
         }
     }
 
@@ -197,7 +220,7 @@ public class SimpleRobotRules extends BaseRobotRules {
         }
     }
 
-    private boolean ruleMatches(String text, String pattern) {
+    private int ruleMatches(String text, String pattern) {
         int patternPos = 0;
         int textPos = 0;
 
@@ -223,7 +246,7 @@ public class SimpleRobotRules extends BaseRobotRules {
                 patternPos += 1;
                 if (patternPos >= patternEnd) {
                     // Pattern ends with '*', we're all good.
-                    return true;
+                    return pattern.length();
                 }
 
                 // TODO - don't worry about having two '*' in a row?
@@ -255,14 +278,14 @@ public class SimpleRobotRules extends BaseRobotRules {
 
                 // If we matched, we're all set, otherwise we failed
                 if (!matched) {
-                    return false;
+                    return -1;
                 }
             } else {
                 // See if the pattern from patternPos to wildcardPos matches the
                 // text starting at textPos
                 while ((patternPos < wildcardPos) && (textPos < textEnd)) {
                     if (text.charAt(textPos++) != pattern.charAt(patternPos++)) {
-                        return false;
+                        return -1;
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


src/test/resources/robots/conflicting-rules.txt
```diff
@@ -0,0 +1,16 @@
+User-Agent: testy
+Disallow: /foo
+
+User-Agent: curl
+Disallow: /
+
+User-Agent: wget
+Allow: /
+Disallow: /*
```
sebastian-nagel, line 9: The [Google spec](https://developers.google.com/search/docs/advanced/robots/robots_txt#url-matching-based-on-path-values) lists `/*` as equivalent of `/`. Of two equivalent rules the less restrictive is chosen.

src/test/java/crawlercommons/robots/SimpleRobotRulesParserTest.java
```diff
@@ -713,6 +714,30 @@ void testManyUserAgents() throws Exception {
...
+        assertFalse(rules.isAllowed("http://domain.com/tmp/bar/bar.txt"), "merge-rules");
+        assertTrue(rules.isAllowed("http://domain.com/chk/bar/bar.txt"), "merge-rules");
+    }
+
+    @Test
+    void testConflictingRules() throws Exception {
+        BaseRobotRules rules = createRobotRules("wget", readFile("/robots/conflicting-rules.txt"));
+
+        assertTrue(rules.isAllowed("http://domain.com/"), "conflicting-rules");
+        assertFalse(rules.isAllowed("http://domain.com/foo/bar"), "conflicting-rules");
```
sebastian-nagel, line 730: See comment about `/*` in `/robots/conflicting-rules.txt`: If `/*` is equivalent to `/` than `/foo/bar` would be allowed?

sebastian-nagel (changes requested):
Thanks, @yoeduardoj, a very valuable contribution! Sorry, that it took quite a while to look into it. The PR lodoks good and I very appreciate the unit tests. And yes, it's definitely necessary to conform with the RFC.

What do you think about the precedence of rules? See my inline comments. I tried to tackle it in 51bc4aa.

Note: I've started work to port of the "official" RFC unit tests (see #245). The merging of multiple blocks needs a change how we match user agent names (see also #192). The latter could be crucial in combination with block merging.

src/main/java/crawlercommons/robots/SimpleRobotRules.java
```diff
@@ -277,7 +288,11 @@ private boolean ruleMatches(String text, String pattern) {
         // We're at the end, so we have a match if the pattern was completely
         // consumed, and either we consumed all the text or we didn't have to
         // match it all (no '$' at end of the pattern)
-        return (patternPos == patternEnd) && ((textPos == textEnd) || !containsEndChar);
+        if ((patternPos == patternEnd) && ((textPos == textEnd) || !containsEndChar)) {
+            return textPos;
```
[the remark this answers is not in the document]
yoeduardoj: Technically if you look carefully at how its done by google, they use pattern length, regardless of match positioning. I've changed to using that approach after addressing other issues.

src/test/resources/robots/conflicting-rules.txt
```diff
@@ -0,0 +1,16 @@
+User-Agent: testy
```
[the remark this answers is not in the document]
yoeduardoj: Renamed.

src/test/resources/robots/conflicting-rules.txt
yoeduardoj: Well, they say that, but their code doesn't actually do that at all. I tested this with some examples in the wild, and my own concocted examples. Given / and /*, they'll assuming /* is of higher priority (because it also matches, and pattern length is greater).

src/test/java/crawlercommons/robots/SimpleRobotRulesParserTest.java
yoeduardoj: Correct, if we go by Google's documentation. Their implementation however, does not. 

One thing I found through addressing your comments is that google applies special rules to Rules such as:

```
Allow: /index.html
```

They actually process this twice, once for an exact match, and a second time where they handle it as
```
Allow: /$
```

This is what allows robots.txt such as https://venmo.com/robots.txt to work as intended (through not by the spec). I've adjusted this PR to take this into account as well.

src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
```diff
@@ -625,17 +612,22 @@ private void handleUserAgent(ParseState state, RobotToken token) {
                 agentName = agentName.trim().toLowerCase(Locale.ROOT);
                 if (agentName.isEmpty()) {
                     // Ignore empty names
-                } else if (agentName.equals("*") && !state.isMatchedWildcard()) {
+                } else if (agentName.equals("*") && !state.isMatchedRealName()) {
                     state.setMatchedWildcard(true);
                     state.setAddingRules(true);
                 } else {
                     // TODO use longest match as winner
```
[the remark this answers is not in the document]
yoeduardoj: Good question, this is where its matching for user-agent. I've removed the TODO for now. The code does stop at the first successful call to startsWith(agentName).

src/test/resources/robots/conflicting-rules.txt
yoeduardoj: The RFC doesn't specify that `/*` needs to be handled in a special way, so I think this version is more RFC-compliant (and is more consistent with Google-code, rather than their docs)

src/main/java/crawlercommons/robots/SimpleRobotRules.java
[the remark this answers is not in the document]
sebastian-nagel: > they use pattern length

Using `patternPos` could only make a difference for patterns with a trailing `$` or `*`.

> I've changed to using that approach

+1

sebastian-nagel (approved):
Thanks, @yoeduardoj! Could you run `java-formatter:format` finally? In order to avoid that code lines are folded, you could add a trailing `//` to the lines. Otherwise, I'll can apply the code format.

yoeduardoj:
> Thanks, @yoeduardoj! Could you run `java-formatter:format` finally? In order to avoid that code lines are folded, you could add a trailing `//` to the lines. Otherwise, I'll can apply the code format.

Applied the formatter as a separate commit. Let me know if you want me to re-base.

sebastian-nagel:
Thanks, @yoeduardoj! Rebased and merged.

PR #360, opened 2022-02-06
Implement Robots Exclusion Protocol (REP) IETF Draft: port unit tests

sebastian-nagel:
Port unit tests from https://github.com/google/robotstxt : https://github.com/google/robotstxt/blob/master/robots_test.cc
- not all tests pass!
  - some tests might be Google-specific (for example limits on line length in the robots.txt) and not generally applicable
  - but others might indicate bugs (see also #351)
- some TODOs left:
  - tests we have no API endpoints now (we eventually may in the future)
  - tests to test methods used in tests (maybe skip them)

### DIFF
--- a/src/test/java/crawlercommons/robots/GoogleRobotsTxtTest.java
+++ b/src/test/java/crawlercommons/robots/GoogleRobotsTxtTest.java
@@ -45,9 +45,7 @@ public class GoogleRobotsTxtTest {
     }
 
     boolean isValidUserAgentToObey(String userAgent) {
-        // TODO: how crawler-commons should handle this?
-        // There is no API endpoint to verify User-Agent names.
-        return userAgent != null && userAgent.matches("^[a-zA-Z_-]+$");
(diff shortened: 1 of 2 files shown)


sebastian-nagel:
Rebased on top of #362 - still five unit tests fail:
- Unicode characters in allow/disallow directives (opened #389)
- one test needs clarification how non-RFC directives (sitemap, crawl-delay) impact closing of rule blocks (see #390)
- one test includes Google-specifics regarding user-agent matching (asked for clarification, cf. /google/robotstxt#56)
- two tests are marked as Google-specific

ontoport:
Hi there, I would like to use url-frontier. What is the current state? Which version would you recommend to begin with?

ontoport:
Hi there, I would like to use the latest version of url-frontier. But here it seems that further development has stopped. How should I proceed? What is your recommendation?

rzo1 (approved):

PR #362, opened 2022-02-20
[Robots.txt] RFC compliance: matching user-agent names when selecting rule blocks

sebastian-nagel:
- add unit test to verify that the rule with the completely matched user-agent name is selected, and no partial prefix match is preferred (cf. also #192)

### DIFF
--- a/src/main/java/crawlercommons/robots/BaseRobotsParser.java
+++ b/src/main/java/crawlercommons/robots/BaseRobotsParser.java
@@ -17,6 +17,7 @@
 package crawlercommons.robots;
 
 import java.io.Serializable;
+import java.util.Collection;
 
 @SuppressWarnings("serial")
 public abstract class BaseRobotsParser implements Serializable {
@@ -37,6 +38,13 @@ public abstract class BaseRobotsParser implements Serializable {
      * "crawlerbot" as the agent name, because of splitting on spaces,
      * lower-casing, and the prefix match rule.
      * 
+     * @deprecated since 1.4 - replaced by {@link parseContent(String, byte[],
+     *             String, Collection)}. Passing a collection of robot names
+     *             gives users more control how user-agent and robot names are
+     *             matched. Passing a list of names is also more efficient as it
+     *             does not require to split the robot name string again and
+     *             again on every robots.txt file to be parsed.
+     * 
      * @param url
      *            URL that robots.txt content was fetched from. A complete and
      *            valid URL (e.g., https://example.com/robots.txt) is expected.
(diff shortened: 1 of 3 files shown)


rzo1 (approved):

sebastian-nagel:
Thanks, @rzo1! But I should add also add a fix. The (failing) unit tests are just an example that the prefix matching is even somewhat counter-intuitive in addition that it contradicts both the [new](https://datatracker.ietf.org/doc/draft-koster-rep/) and the [old RFC](https://www.robotstxt.org/norobots-rfc.txt).

sebastian-nagel:
Updated the PR...

To summarize the issues around matching user agent names:

1. the new [RFC 9309](https://datatracker.ietf.org/doc/rfc9309/) is pretty clear: both the crawler (robot, user-agent) name and the name in the user-agent line are required to be a single word token (regex `[a-zA-Z_-]+`) and matching is done literally but case-insensitive on the entire word token.

2. the [first RFC proposal](http://www.robotstxt.org/norobots-rfc.txt) only requires that the user-agent token is a substring of the User-Agent line in the robots.txt:

   The robot must obey the first record in /robots.txt that contains a User-Agent line whose value contains the name token of the robot as a substring.
   
3. our current implementation is different from 1 and 2:

   a. the user-agent line is split at `[ \t,]' into tokens
   b. the configured user-agent name is also split into tokens at white space (U+0020)
   c. a user-agent line matches if one of the line tokens (from a) is a prefix of one of the name tokens (from b)
   d. the first matched non-wildcard block of allow/disallow rules is selected

   In short, the current implementation allows that crawler developers lazily use the HTTP User-Agent string also for the robots.txt parser. It does not cover the case when the HTTP User-Agent string is used in the robots.txt.

4. the current robots.txt parser API allows to pass
   a. multiple agent names in a comma-separated string
   b. and/or compound agent names such as "Mozilla Crawlerbot-super 1.0"
      where every white-space separated word is matched

Issues with the current implementation are conflicts when robot names share a common prefix, eg. `googlebot` and `googlebot-news`. But the real trouble (that's #192 is about) starts when both user-agent line and name consist of multiple tokens and include common parts such as "Mozilla".


Decisions to resolve this issues:

1. extend the robots parser API and add a method to pass agent names as a collection following the RFC 9309 with no splitting of the names into words/tokens.

   This still allows for multiple robot names, including varying names which include version string or similar ("webcrawler/3.0").

2. by default user agent names are matched literally but case-insensitive following RFC 9309

3. the old behavior can be restored by calling "setExactUserAgentMatching(false)"

4. the method which splits the robot name into tokens and performs prefix matching is deprecated but kept for backward-compatibility.

5. BaseRobotRulesParser: move the documented details about how user-agent names are matched into SimpleRobotRulesParser. The abstract base class shouldn't already define the behavior without a way to actually enforce it.

sebastian-nagel:
Following the discussion in google/robotstxt#56, the behavior should be changed, so that a robot name "foo" also matches all of:
```
User-agent: Foo/1.2
User-agent: Foo Bar
```
Eventually, make the user-agent matching configurable and implement it in a method which can be overridden if a user is stick to an agent name containing any characters except `[a-zA-Z_-]` - "Mail.Ru", "Go!Zilla", "360Spider", etc.

sebastian-nagel:
A "prefix" match up to the first non-product-token character is now implemented. I.e. the product token "foo" is matched in `User-agent: Foo/1.2`. To handle partial matches of special product tokens ("Go!Zilla", etc.) the method "userAgentProductTokenPartialMatch" can be overridden.

jnioche:
LGTM @sebastian-nagel, nice one! Feel free to merge

rzo1 (approved):

sebastian-nagel:
Thanks, @jnioche and @rzo1!

PR #365, opened 2022-02-21
Enhance GitHub actions with actions/cache for caching maven artifacts

rzo1:
# What does this PR do?

- Use actions/cache to prevent downloading 3rd party artifacts again and again (they are not changing that often)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,5 +25,15 @@ jobs:
           java-version: ${{ matrix.java }}
           cache: 'maven'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -25,5 +25,15 @@ jobs:
           java-version: ${{ matrix.java }}
           cache: 'maven'
```
valfirst, line 26: the changes below are not needed, since caching is enabled at this line, see: https://github.com/actions/setup-java#caching-packages-dependencies

PR #371, opened 2022-02-25
Not allow doctype when parsing sitemap

kennethwong-hc:
### Description of Changes
- Added feature http://apache.org/xml/features/disallow-doctype-decl when creating SAXParser for site maps.
- Allow exempt if particular property has been set.

### Reference
- https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#java

### Risks & Impacts
- Added option `allowDocTypeDefinitions `to allow DTD in `SiteMapParser `(disabled by default)
- Added new feature in site map parser if DTD is not allowed
- Will not parse any sitemaps with DOCTYPE definition, if `allowDocTypeDefinitions `isn't set as true.

### Security
- Prevent possible DTD vulnerability when parsing site maps

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
+++ b/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
@@ -595,7 +595,9 @@ public class SiteMapParser {
             factory.setFeature("http://xml.org/sax/features/external-general-entities", false);
             factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
             factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
-            factory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+            if (!"true".equalsIgnoreCase(System.getProperty("crawler-commons.sitemap.allowDocTypes"))) {
+                factory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+            }
         } catch (Exception e) {
             throw new RuntimeException("Failed to configure XML parser: " + e.toString());
         }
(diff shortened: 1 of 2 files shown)


src/test/java/crawlercommons/sitemaps/SiteMapParserTest.java
```diff
@@ -114,6 +114,7 @@ public void testSitemapIndex() throws UnknownFormatException, IOException {
 
     @Test
     public void testSitemapXXE() throws UnknownFormatException, IOException {
```
sebastian-nagel, line 116: UnknownFormatException isn't thrown because it's handled via assertions.

src/main/java/crawlercommons/sitemaps/SiteMapParser.java
```diff
@@ -595,6 +595,9 @@ protected AbstractSiteMap processXml(URL sitemapUrl, InputSource is) throws Unkn
             factory.setFeature("http://xml.org/sax/features/external-general-entities", false);
             factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
             factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
+            if (!"true".equalsIgnoreCase(System.getProperty("crawler-commons.sitemap.allowDocTypes"))) {
```
sebastian-nagel, line 598: [Boolean.getBoolean(...)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Boolean.html#getBoolean(java.lang.String)) would be shorter and does exactly the same.

Until know properties are only used in the class SiteMapTester. To keep the SiteMapParser consistent: what about adding a setter method `setAllowDocTypeDefinitions(boolean)` which is used in SiteMapTester and unit tests. This would allow to switch the support for DTDs on and off programmatically.

sebastian-nagel (changes requested):
Hi @kennethwong-hc, thanks! Let me know whether I shall take care to change the way the option is set.

Any other comments? Is it ok, to disable DTDs by default. This might break the parsing of one or the other sitemap. But I'd expect only few.

kennethwong-hc:
Hi @sebastian-nagel , thanks for your comment. I think disabling DTD by default is better, as it is a vulnerability, and providing an option to allow DTD should not break anything that work now. 
An attribute has been added in `SiteMapParser`, so it would not be a system setting, but per parser

sebastian-nagel (approved):
Thanks, @kennethwong-hc! I'm going to merge your PR (I'll add a line to the change log and apply the code formatting (`mvn java-formatter:format`).

PR #380, opened 2022-08-08
Javadoc: ensure Javascript search is working, fixes #378

### LINKED ISSUE #378: Javadoc: ensure Javascript search is working
Newer Java versions create a nice Javascript-backed search box on the Javadocs. Because crawler-commons does not use Java modules, querying via this search box does not work and leads to 404 results with a `/undefined/` path element in the URL.

- cf. https://stackoverflow.com/questions/52326318/maven-javadoc-search-redirects-to-undefined-url

- seen during the release of 1.3
  - work-around possible by adding the following lines to the file search.js (see 7291f45):
    ```javascript
       getURLPrefix = function(ui) {
         return "";
       };
    ```

sebastian-nagel:
Only for JDK / Java 11, pass the option `--no-module-directories` when building javadocs:
- the issue #378 only affects Javadoc artifacts built with JDK 11 (but not JDK 8 and 17 - I didn't test other versions)
- the option `--no-module-directories` breaks Javadoc builds on JDK 17, so it's made sure it's only called for JDK 11
- according to [JDK-8215291](https://bugs.openjdk.org/browse/JDK-8215291) the issue is fixed for JDK 12 and later

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -247,6 +247,28 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-javadoc-plugin</artifactId>
+				<version>${maven-javadoc-plugin.version}</version>
+				<configuration>
+					<links>
+						<link>https://docs.oracle.com/javase/8/docs/api/</link>
+					</links>
+					<source>${javac.src.version}</source>
+					<additionalJOptions>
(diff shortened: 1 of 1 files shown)


rzo1 (approved):

aecio (approved):

sebastian-nagel:
Thanks, @rzo1 and @aecio! Going to merge...

PR #412, opened 2023-05-23
Generate JaCoCo reports when testing

### LINKED ISSUE #409: Add test coverage reports with JaCoCo
and export to [Coveralls](https://coveralls.io/) perhaps?

Knowing the area of the library for which we are lacking tests would be quite useful I think.

Any thoughts on this?

jnioche:
Fixes #409

image

I have put a secret COVERALLS_REPO_TOKEN in the GH config, just in case @rzo1 wants to send the JaCoco results to [Coveralls](https://coveralls.io/github/crawler-commons/crawler-commons) ;-)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,6 +158,26 @@
 
 		<plugins>
 			<plugin>
+				<groupId>org.jacoco</groupId>
+				<artifactId>jacoco-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


sebastian-nagel (approved):
Great! Really useful coverage report (checked the HTML version in target/site/jacoco/).

PR #413, opened 2023-05-23
[Robots.txt] SimpleRobotRulesParser main to use the new API method

sebastian-nagel:
... without splitting the agent name into tokens. Simply pass a single user-agent "product token" as per RFC 9309. See #362 for details regarding user-agent name matching.


jnioche:
LGTM

PR #421, opened 2023-06-13
[BasicNormalizer] Query parameters normalization in BasicURLNormalizer

### LINKED ISSUE #308: [BasicNormalizer] Query parameters normalization in BasicURLNormalizer

sebastian-nagel:
- add unit test to prove that an empty query is removed, closes #308

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,7 @@
 Crawler-Commons Change Log
 
 Current Development 1.4-SNAPSHOT (yyyy-mm-dd)
+- [BasicNormalizer] Query parameters normalization in BasicURLNormalizer (aecio, sebastian-nagel) #308
 - [Robots.txt] Handle allow/disallow directives containing unescaped Unicode characters (sebastian-nagel, Richard Zowalla, aecio) #389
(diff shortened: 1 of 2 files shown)


rzo1 (approved):

sebastian-nagel:
Thanks, @rzo1!

PR #424, opened 2023-06-13
[Robots.txt] Empty disallow statement not to clear other rules, fixes #422

### LINKED ISSUE #422: [Robots.txt] Empty disallow statement not to clear other rules
In the current implementation of SimpleRobotRulesParser ([line 819](https://github.com/crawler-commons/crawler-commons/blob/0656e9c56137bc66bdd47238f2b6b45c36c30280/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java#L819)), an empty disallow statement (`Disallow: ` ) clears all rules read so far. It should not.

- the [initial robots.txt RFC draft](https://www.robotstxt.org/robotstxt.html) didn't include the `Allow:` statement and the empty disallow statement was required to express that all paths are allowed.

- the [formal syntax of RFC 9309](https://www.rfc-editor.org/rfc/rfc9309.html#section-2.2) allows empty patterns in allow/disallow statements. However, because longer rules take precedence, empty patterns would be matched last.

The issue was observed with [this robots.txt](https://agriexotic.com/robots.txt). Here the empty disallow statement in the second rule group, clears the already parsed rules. Both groups are addressed to the wildcard user-agent and therefore merged.

sebastian-nagel:
Also verified that Google's reference parser behaves the same:

```sh
$> cat empty-disallow-robots.txt 
User-agent: *
Disallow: /disallowed/
Allow: /allowed/
Disallow: 

$> for path in "" / /allowed/page.html /disallowed/page.html; do
    robots empty-disallow-robots.txt any https://www.example.com$path;
done
user-agent 'any' with URI 'https://www.example.com': ALLOWED
user-agent 'any' with URI 'https://www.example.com/': ALLOWED
user-agent 'any' with URI 'https://www.example.com/allowed/page.html': ALLOWED
user-agent 'any' with URI 'https://www.example.com/disallowed/page.html': DISALLOWED
```

### DIFF
--- a/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
+++ b/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
@@ -815,8 +815,20 @@ public class SimpleRobotRulesParser extends BaseRobotsParser {
         try {
             path = normalizePathDirective(path);
             if (path.length() == 0) {
-                // Disallow: <nothing> => allow all.
-                state.clearRules();
+                /*
+                 * "Disallow: <nothing>" meant "allow all" in the 1996 REP RFC
+                 * draft because there was no "Allow" directive.
+                 * 
+                 * RFC 9309 allows empty patterns in the formal syntax
+                 * (https://www.rfc-editor.org/rfc/rfc9309.html#section-2.2). No
+                 * specific handling of empty patterns is defined, as shortest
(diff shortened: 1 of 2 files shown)


jnioche (approved):

jnioche:
Thanks @sebastian-nagel

PR #425, opened 2023-06-13
[Robots.txt] Add more spelling variants and typos of robots.txt directives

sebastian-nagel:
- found in Google's RFC 9309 reference parser (see [robots.cc](https://github.com/google/robotstxt/blob/master/robots.cc#L690))
- and in real-world robots.txt files (Common Crawl)
- if we accept lines starting with `http:` as sitemap directives we should nowadays also accept `https:` as such

Add Javadoc comments for some of the robots.txt extension directives.


jnioche:
thanks @sebastian-nagel

PR #426, opened 2023-06-15
[Robots.txt] Document effect of rules merging in combination with multiple agent names, fixes #423

### LINKED ISSUE #423: [Robots.txt] Document effect of rules merging in combination with multiple agent names
[RFC 9309 section 2.2.1](https://www.rfc-editor.org/rfc/rfc9309.html#name-the-user-agent-line) requires that groups of rules targeted to the same user-agent are merged. See also [Google's spec, "Grouping of rules"](https://developers.google.com/search/docs/crawling-indexing/robots/robots_txt#grouping-of-rules). With #351 SimpleRobotRulesParser implements the merge of rule groups.

However, RFC 9309 assumes that a crawler uses a single name (or "product token"). In contrary, our API (defined in BaseRobotsParser) allows to select rules using a collection of names. The current implementation (1.4 snapshot) merges all groups matched by any of the agent names. This should be documented including all side effects with clear recommendations, e.g., never pass the wildcard user-agent `*` along with other user-agent names. This was also a requirement before (1.3 and earlier), cf. [NUTCH-1715](https://issues.apache.org/jira/browse/NUTCH-1715).

Note that it's not uncommon that multiple agents are supported:
- Google's robots.txt parser API (the RFC 9309 reference implementation) also accepts a list/vector of agent names ([robots.h, line 116](https://github.com/google/robotstxt/blob/455b1583103d13ad88fe526bc058d6b9f3309215/robots.h#L116)). See google/robotstxt#26 for a the specific use case.
- Applebot looks for rules addressing - in this order - "Applebot", "Googlebot", "*" ([1](https://support.apple.com/en-us/HT204683), [2](http://www.thesempost.com/apples-applebot-follows-googlebots-instructions-in-robots-txt-files/)).

For now, we should document the behavior that the rule groups of all matched agent names are merged. We might later implement other strategies (eg. Apple's) to select a particular rule group.

sebastian-nagel:
Update documentation related to RFC 9309 compliance
- document effect of rules merging in combination with multiple agent names (#423)
- document that rules addressed to the wildcard agent are followed if none of the passed agent names matches - without any need to pass the wildcard agent name as one of the agent names
- complete robots.txt parser documentation
- use @inheritDoc to avoid duplicated documentation
- strip doc strings where inherited automatically per @Override annotations

### DIFF
--- a/src/main/java/crawlercommons/robots/BaseRobotRules.java
+++ b/src/main/java/crawlercommons/robots/BaseRobotRules.java
@@ -22,8 +22,11 @@ import java.util.LinkedHashSet;
 import java.util.List;
 
 /**
- * Result from parsing a single robots.txt file - which means we get a set of
- * rules, and a crawl-delay.
+ * Result from parsing a single robots.txt file – a set of allow/disallow rules
+ * to check whether a given URL is allowed, and optionally a <a href=
+ * "https://en.wikipedia.org/wiki/Robots.txt#Crawl-delay_directive"
+ * >Crawl-delay</a> and <a
+ * href="https://www.sitemaps.org/protocol.html#submit_robots">Sitemap</a> URLs.
  */
 @SuppressWarnings("serial")
 public abstract class BaseRobotRules implements Serializable {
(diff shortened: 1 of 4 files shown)


sebastian-nagel:
Thanks, @rzo1!

PR #428, opened 2023-06-16
[Robots.txt] SimpleRobotRulesParser main() to follow five redirects

sebastian-nagel:
Follow "five consecutive redirects" when fetching robots.txt over HTTP as required by [RFC 9309](https://www.rfc-editor.org/rfc/rfc9309.html#name-redirects).


jnioche (approved):
LGTM

PR #430, opened 2023-06-16
[Robots.txt] Close groups of rules as defined in RFC 9309

### LINKED ISSUE #114: Handle robots.txt with missing sections (and implicit master rules)
The robots.txt file at http://www.scotsman.com/robots.txt has a number of issues...
1. There's no blank line between rule sections
2. The "*" user agent rule section is obviously intended to be used by the explicit (e.g. "bingbot") rule sections, even though it technically won't be.
3. This isn't really an issue, but we should make sure we process all sitemap refs, as there is one in the middle of the general rule block, and a bunch more at the end.

### LINKED ISSUE #390: [Robots.txt] Clarify behavior when to close blocks of multiple user-agents
The RFC reference parser includes a unit test to verify that sitemap and other directives not covered by the RFC are just ignored and do not close blocks of rules. See [robots_test.cc, line 126](https://github.com/google/robotstxt/blob/455b1583103d13ad88fe526bc058d6b9f3309215/robots_test.cc#L126) expecting that BarBot is disallowed given:
```
User-agent: BarBot
Sitemap: https://foo.bar/sitemap
User-agent: *
Disallow: /
```

However, this interpretation might contradict the actual usage of certain non-RFC directives, see for example https://www.skyrock.com/robots.txt:
```
User-agent: *
Disallow: /cache/
Disallow: /demo/
...

User-agent: Twitterbot
Allow: /img/

User-agent: bingbot
Crawl-delay: 30

User-agent: 007ac9
User-agent: AhrefsBot
User-agent: LinqiaBot
User-agent: LinqiaRSSBot
User-agent: SMTBot
User-agent: HaosouSpider
User-agent: OpenindexSpider
User-agent: PetalBot
User-agent: serpstatbot
Disallow: / 

...
```

It's likely not not intended to exclude bingbot entirely - in fact, it wouldn't make sense to define a crawl-delay if a user-agent is entirely excluded. See also the discussion in google/robotstxt#51.

sebastian-nagel:
This PR addresses:
- Clarify behavior when to close blocks of multiple user-agents (resolves #390)
- Handle robots.txt with missing sections (and implicit master rules (resolves #114)

Changes:
- rule groups are closed as specified in RFC 9309: on a user-agent line, if at least one allow/disallow line was read before
- that is, Sitemap, Crawl-delay and other directives not covered by RFC 9309 do not finish a rule group
- the Crawl-delay is set independently from grouping:
  - it's set for a given agent or the wildcard agent
  - but a value defined for a specific agent is never set using a value defined for the wildcard agent
- unit tests are updated or added accordingly
  - for changed tests it's verified, that the RFC reference parser ([google/robotstxt](https://github.com/google/robotstxt)) behaves the same

This PR should unblock #245 (all unit tests added there should now pass).

### DIFF
--- a/src/main/java/crawlercommons/robots/BaseRobotRules.java
+++ b/src/main/java/crawlercommons/robots/BaseRobotRules.java
@@ -47,10 +47,20 @@ public abstract class BaseRobotRules implements Serializable {
         _sitemaps = new LinkedHashSet<>();
     }
 
+    /**
+     * Get Crawl-delay (in milliseconds)
+     * 
+     * @return Crawl-delay defined in the robots.txt for the given agent name,
+     *         or {@link UNSET_CRAWL_DELAY} if not defined.
+     */
     public long getCrawlDelay() {
         return _crawlDelay;
     }
 
+    /**
+     * @param crawlDelay
+     *            Crawl-Delay in milliseconds
+     */
     public void setCrawlDelay(long crawlDelay) {
         _crawlDelay = crawlDelay;
     }
--- a/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
+++ b/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
@@ -204,6 +204,12 @@ public class SimpleRobotRulesParser extends BaseRobotsParser {
          * something else has been found.
          */
         private boolean _finishedAgentFields;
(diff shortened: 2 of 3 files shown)


sebastian-nagel:
Thanks for the review, @jnioche!

Issue #145, opened 2017-02-06
[Sitemaps] limit on "bad url" log messages


sebastian-nagel:
If a sitemap is erroneously detected as plain-text sitemap (cf. #144), SiteMapParser may report all or most of the file content as "bad url". This may result
- either in many log messages:
```
2016-12-22 13:55:26.628 c.s.SiteMapParser [WARN] Bad url: [<?xml version="1.0" encoding="UTF-8"?>]
2016-12-22 13:55:26.628 c.s.SiteMapParser [WARN] Bad url: [<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:news="http://www.google.com/schemas/sitemap-news/0.9">]
2016-12-22 13:55:26.628 c.s.SiteMapParser [WARN] Bad url: [                <url>]
2016-12-22 13:55:26.628 c.s.SiteMapParser [WARN] Bad url: [        <loc>http://www.azonline.de/Sport/Fussball/1.-Bundesliga/2640159-Hamburger-SV-Bruchhagen-Lasse-mich-bei-der-Sportchef-Suche-nicht-hetzen</loc>]
2016-12-22 13:55:26.628 c.s.SiteMapParser [WARN] Bad url: [        <news:news>]
2016-12-22 13:55:26.628 c.s.SiteMapParser [WARN] Bad url: [            <news:publication>]
2016-12-22 13:55:26.628 c.s.SiteMapParser [WARN] Bad url: [                <news:name>Allgemeine Zeitung</news:name>]
2016-12-22 13:55:26.628 c.s.SiteMapParser [WARN] Bad url: [                <news:language>ger</news:language>]
2016-12-22 13:55:26.628 c.s.SiteMapParser [WARN] Bad url: [            </news:publication>]
2016-12-22 13:55:26.628 c.s.SiteMapParser [WARN] Bad url: [            <news:publication_date>2016-12-22T14:52:00Z</news:publication_date>]
2016-12-22 13:55:26.628 c.s.SiteMapParser [WARN] Bad url: [            <news:title>Hamburger SV : Bruchhagen: Lasse mich bei der Sportchef-Suche nicht hetzen</news:title>]
2016-12-22 13:55:26.628 c.s.SiteMapParser [WARN] Bad url: [        </news:news>]
2016-12-22 13:55:26.628 c.s.SiteMapParser [WARN] Bad url: [    </url>]
[...]

PR #457, opened 2024-05-14
Improve logging done by SiteMapParser

valfirst:
- increase log levels to raise importance of certain messages
- optimize performance by using parameterized log messages

### DIFF
--- a/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
+++ b/src/main/java/crawlercommons/sitemaps/SiteMapParser.java
@@ -412,7 +412,7 @@ public class SiteMapParser {
      */
     public void walkSiteMap(URL onlineSitemapUrl, Consumer<SiteMapURL> action) throws UnknownFormatException, IOException {
         if (onlineSitemapUrl == null || action == null) {
-            LOG.debug("Got null sitemap URL and/or action, stopping traversal");
+            LOG.info("Got null sitemap URL and/or action, stopping traversal");
             return;
         }
         walkSiteMap(parseSiteMap(onlineSitemapUrl), action);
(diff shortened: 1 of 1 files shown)


sebastian-nagel:
Hi @valfirst, thanks for the PR! Could you share the objective why to increase the log level? I can understand that showing all bad lines or URLs without setting the level to DEBUG helps to detect potential format errors in sitemaps and similar. However, there was reason to decrease the log level (see #145) - this was some time ago and the MIME type detection of the sitemap parser was less precise at this time. It's about to avoid that any erroneously "announced" sitemap (for example, a HTML file referenced in the robots.txt) floods the log file. I will check how this problem affects the sitemap parser with real-world sitemap links from robots.txt files. Eventually, we could also log only the first N "bad URLs" per sitemap.

valfirst:
@sebastian-nagel In general your guess is correct. In my case `sitemap.xml` was valid, but contained a set of invalid URLs, the messages about them were not logged, because we didn't enable `debug` level globally. I checked source code and enabled `debug` level for `crawlercommons.sitemaps.SiteMapParser`, but it resulted in number of not so important messages. I thought it could be useful to filter logged information and I changed logging in `SiteMapParser` based on my understanding of its importance.

I've not faced with cases like described in #145 and now I see the reasoning behind it.

>  Eventually, we could also log only the first N "bad URLs" per sitemap.

If you decide to go with this approach, I can help implementing this solution.

sebastian-nagel:
Hi @valfirst,
 
 > the MIME type detection of the sitemap parser was less precise at this time.
 > [...] check how this problem affects the sitemap parser with real-world sitemap links from robots.txt files.
 
Done. The detection of the MIME type of sitemaps is quite precise after we implemented a [simple MIME detector](https://github.com/crawler-commons/crawler-commons/blob/master/src/main/java/crawlercommons/mimetypes/MimeTypeDetector.java) (#198) which supports only the very limited set of "sitemap MIME types". I found *no* HTML sitemap which was erroneously detected as "text/plain". Consequently, there were *no* erroneous "Bad url: ..." log messages.

> >  Eventually, we could also log only the first N "bad URLs" per sitemap.

> If you decide to go with this approach, I can help implementing this solution.

Thanks! It's on you. It does not harm but it seems not necessary anymore.

valfirst:
> Thanks! It's on you. It does not harm but it seems not necessary anymore.

let's skip it, I like YAGNI principle

sebastian-nagel (approved):

sebastian-nagel:
Thanks, @valfirst!

PR #478, opened 2024-10-31
[Domains] Improve representation of public suffix match results by class EffectiveTLD

sebastian-nagel:
The class [EffectiveTldFinder.EffectiveTLD](https://crawler-commons.github.io/crawler-commons/1.4/crawlercommons/domains/EffectiveTldFinder.EffectiveTLD.html) is used for two purposes:
1. it represents one line (one suffix) of the public suffix list, i.e. the suffix string and its properties: whether it's wildcard suffix, an exception, or is part of the private section of the list
2. it holds the result of a suffix match returned by [getEffectiveTLD(hostname)](https://crawler-commons.github.io/crawler-commons/1.4/crawlercommons/domains/EffectiveTldFinder.html#getEffectiveTLD(java.lang.String))

Due to this dual use, there's some confusion what EffectiveTLD shall contain as a match result.

Let's look at two problematic matches:

- matching a wildcard suffix, `*.bd` in the public suffix list:
   ```
   // bd : https://en.wikipedia.org/wiki/.bd
   *.bd
   ```
   The match result does not state that it's a wildcard suffix. As `domain` it shows the literally matched suffix, but it does not provide the original line from the public suffix list:
   ```sh
   echo abc.def.bd | java -cp ... crawlercommons.domains.EffectiveTldFinder -etld -excludePrivate -strict
   abc.def.bd      [domain=def.bd,wild=false,exception=false,private=false]
   ```
- in case of IDNs, only the ASCII version of the suffix is given, the literally matched suffix is not part of the result:
   ```
   中国政府网.政务   [domain=xn--zfr164b,wild=false,exception=false,private=false]
   ```

This PR addresses these two points – the results are now:
```
abc.def.bd      [domain=def.bd,suffix=*.bd,idn=null,wildcard=true,exception=false,private=false]
中国政府网.政务   [domain=政务,suffix=xn--zfr164b,idn=政务,wildcard=false,exception=false,private=false]
```

In detail: 
- ensure the flag `wildcard` (and all other flags) are transferred from the ETLD representing the public suffix line to that holding the match result
- added the method `EffectiveTLD.getSuffix()` which returns the suffix as specified in the public suffix list, e.g. `*.bd`
- `getDomain()` still returns the literally matched suffix (ETLD), also for Unicode suffixes (IDNs)
- add method EffectiveTLD.getUnicodeDomain() which - in case of an IDN - returns the Unicode representation of the suffix
- rename `wild` to `wildcard` and `isWild()` to `isWildcard()`

### DIFF
--- a/src/main/java/crawlercommons/domains/EffectiveTldFinder.java
+++ b/src/main/java/crawlercommons/domains/EffectiveTldFinder.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2016 Crawler-Commons
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


rzo1 (approved):

sebastian-nagel:
Thanks for the review, @rzo1!

PR #493, opened 2025-03-13
[Robots] Document that URLs not properly normalized may not be matched by robots.txt parser, resolves #492

### LINKED ISSUE #492: [Robots] Document that URLs not properly normalized may not be matched by robots.txt parser
If a URL is not properly normalized, it's path may not be matched by SimpleRobotRulesParser:

```
$> cat robots.txt
User-agent: *
Disallow: /forbiddenpath/

$> java -cp ... crawlercommons.robots.SimpleRobotRulesParser \
    file:$PWD/robots.txt \
    anyagent \
    https://www.example.org/ \
    https://www.example.org/forbiddenpath/file1.html \
    https://www.example.org//forbiddenpath/file2.html

Checking URLs:
allowed         https://www.example.org/
forbidden       https://www.example.org/forbiddenpath/file1.html
allowed         https://www.example.org//forbiddenpath/file2.html
```

This is hardly an issue of the robots.txt parser
- other parsers including [Google's parser](https://github.com/google/robotstxt) behave identical
- [RFC 9309](https://www.rfc-editor.org/rfc/rfc9309.html) does not say anything about URL canonicalization, but only defines how percent-encodings must be handled.

Nevertheless, we should document the fact that URL paths not properly normalized may not be matched.
- a short example how to use BasicURLNormalizer might help
- also a few examples (with unit tests) were matching fails: double slashes `//`, dot segments, etc.

sebastian-nagel:
- add unit test to proof that URLs not normalized may fail to match and that after normalization the URLs match
- add comment to Javadoc about the requirement to normalize URLs

### DIFF
--- a/src/main/java/crawlercommons/robots/SimpleRobotRules.java
+++ b/src/main/java/crawlercommons/robots/SimpleRobotRules.java
@@ -149,6 +149,14 @@ public class SimpleRobotRules extends BaseRobotRules {
         _rules.clear();
     }
 
+    /**
+     * Add an allow/disallow rule to the ruleset
+     * 
+     * @param prefix
+     *            path prefix or pattern
+     * @param allow
+     *            whether to allow the URLs matching the prefix or pattern
+     */
     public void addRule(String prefix, boolean allow) {
         // Convert old-style case of disallow: <nothing>
         // into new allow: <nothing>.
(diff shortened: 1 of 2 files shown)


kkrugler (approved):
lgtm

sebastian-nagel:
Thanks for the review, @kkrugler!

Issue #453, opened 2023-12-07
[Robots.txt] Incorrect robots txt result for uppercase user agents


teammakdi:
Example for forbes.com robots txt

https://www.forbes.com/robots.txt

They have blocked all paths for `GPTBot`

```
User-agent: GPTBot
Disallow: /
```

However for url `https://www.forbes.com/test`

```java
public boolean canCrawl(String url, String userAgent, String robotsBody)
            throws MalformedURLException {
        SimpleRobotRulesParser robotParser = new SimpleRobotRulesParser();
        robotParser.setExactUserAgentMatching(false);
        BaseRobotRules robotRules =
                robotParser.parseContent(
                        "https://www.forbes.com/robots.txt",
                        robotsBody.getBytes(StandardCharsets.UTF_8),
                        "text/plain",
                        Collections.singletonList(userAgent));
        return robotRules.isAllowed(url);
    }
```

returns `true`

The function is called as

```java
boolean canCrawl = canCrawl("https://www.forbes.com/test", "GPTBot", "<robots body>");
```

Verified this behaviour with https://github.com/samclarke/robots-parser and https://github.com/google/robotstxt, they both return `false` which seems correct.

If the UA would have been `gptbot` in robots.txt, it would return false for both `GPTBot` and `gptbot` UA. There seems to be some case sensitive check in the code base.

jnioche:
> But maybe we can do it: check first whether the user-agent name(s) are already lower-case, if not log it and create a new collection with lower-cased user-agent names. @jnioche, what do you think?

we could throw an exception if the tokens don't follow RFC9309 and `exactUserAgentMatching` is true - which would make it impossible to ignore.

If `exactUserAgentMatching` is false then yes, simply log a warning and lower case if needed.

teammakdi:
Hey @jnioche @sebastian-nagel 

Shall I create a PR for the same?

PR #500, opened 2025-04-24
[Robots.txt] Incorrect robots txt result for uppercase user agents

sebastian-nagel:
Address #453:
- throw IllegalArgumentException with exact user-agent matching and upper-case or wildcard user-agent name
- provide method to sanitize (lower-case and remove wild-card) user-agent names
- trivial code cleanup logging and Javadoc

> we could throw an exception if the tokens don't follow RFC9309 and exactUserAgentMatching is true

- this is now implemented but only for upper-case user-agent names
- SimpleRobotRulesParser can handle to some extend user-agent names not following RFC 9309, we even have unit tests for them


rzo1:
Needs a rebase, I guess.

rzo1 (approved):

src/test/java/crawlercommons/robots/SimpleRobotRulesParserTest.java
```diff
@@ -16,6 +16,7 @@
 
 package crawlercommons.robots;
 
+import org.junit.jupiter.api.Assertions;
```
aecio, line 19: Minor: This import seems unnecessary since a static import is already present below (line 37).

src/test/java/crawlercommons/robots/SimpleRobotRulesParserTest.java
```diff
@@ -1383,6 +1384,24 @@ void testAPIemptyUserAgentList() {
...
     }
 
+    @Test
+    void testAPIsanitizeUserAgentList() throws Exception {
+        byte[] robotstxt = readFile("/robots/rfc9309-example-simple-robots.txt");
+
+        // verify that an exception is thrown if the provided user-agent tokens
+        // include a invalid elements
+        Assertions.assertThrows(IllegalArgumentException.class, () -> createRobotRules(Set.of("FOOBOT"), robotstxt, true));
+        Assertions.assertThrows(IllegalArgumentException.class, () -> createRobotRules(Set.of("*"), robotstxt, true));
```
aecio, line 1394: Maybe `Asssertion.` can be removed due to the static import?

aecio (approved):
LGTM> I left only a few minor comments.

PR #503, opened 2025-05-17
Remove class utils.Strings

sebastian-nagel:
The availability of [java.lang.String.isBlank](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#isBlank()) in JDK 11 obsoletes the method utils.Strings.isBlank which is the only method implemented in the class utils.Strings. Consequently it can be removed.


rzo1 (approved):

sebastian-nagel:
Thanks, @rzo1!

Issue #522, opened 2025-07-09
Java 20 deprecates all constructors of the class java.net.URL


sebastian-nagel:
Java 20 deprecates all constructors of the class java.net.URL, see:
- <https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/net/URL.html#constructor-deprecation>
- <https://bugs.openjdk.org/browse/JDK-8294241>

The rationales behind the deprecation are comprehensible. Everybody working on a Java web crawler which uses the URL class (not URI) likely stumbled upon one of the limitations and caveats. In addition, to the validation issues described in the deprecation, there are:
- the `equals` and `hashCode` methods doing a DNS lookup
- dealing with URLStreamHandler if the scheme/protocol is not http(s), file or ftp, e.g., smb://
- a bug with relative URLs, see [NUTCH-797](https://issues.apache.org/jira/browse/NUTCH-797). Side note: StormCrawler uses the same solution.

For now, it's a deprecation without deadline. Java 24 still includes the deprecated constructors.

Nevertheless, as a common core component for multiple Java web crawlers, we might want to
1. avoid the deprecated constructors in our code
2. additionally provide methods taking strings and/or URIs as arguments, so that web crawler implementations are not forced to convert from URIs to URLs when using our methods.

PR #526, opened 2025-07-15
Add URLUtils class for URL resolution functionality

### LINKED ISSUE #525: Create UrlUtils class with resolve(URL base, String spec) method
This issue aims to create a UrlUtils class with a resolve(URL base, String spec) method, as discussed in PR #524  The goal is to encapsulate URL resolution logic and use it as a cleaner replacement for deprecated URL constructors.

HamzaElzarw-2022:
This PR introduces a URL utility class with a static resolve(URL base, String spec) method. It encapsulates logic previously implemented with deprecated URL(URL, String) constructors, improving clarity and future maintainability.

The utility is intended to be used for replacing deprecated URL constructors in PR #524.

closes #525

### DIFF
--- a/src/main/java/crawlercommons/utils/URLUtils.java
+++ b/src/main/java/crawlercommons/utils/URLUtils.java
@@ -1,0 +1,23 @@
+package crawlercommons.utils;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+public class URLUtils {
+
+    /**
+     * Resolves a URL against a base URL.
+     *
+     * @param base the base URL
+     * @param spec the URL specification to resolve
+     * @return the resolved URL
+     * @throws MalformedURLException if the URL cannot be resolved
+     */
+    public static URL resolve(URL base, String spec) throws MalformedURLException {
+        try {
+            return base.toURI().resolve(spec).toURL();
+        } catch (Exception e) {
+            throw new MalformedURLException(e.getMessage());
+        }
+    }
+}


src/main/java/crawlercommons/utils/URLUtils.java
```diff
@@ -0,0 +1,23 @@
...
+     * @param base the base URL
+     * @param spec the URL specification to resolve
+     * @return the resolved URL
+     * @throws MalformedURLException if the URL cannot be resolved
+     */
+    public static URL resolve(URL base, String spec) throws MalformedURLException {
+        try {
+            return base.toURI().resolve(spec).toURL();
+        } catch (Exception e) {
+            throw new MalformedURLException(e.getMessage());
```
rzo1, line 20: I would wrap the underyling exception in the MUE, so the caller has an idea what exactly has gone wrong?

HamzaElzarw-2022: Thank you for the suggestion, I’ll update the code accordingly.

I also wanted to ask: would it be better to catch the specific exception types like URISyntaxException, IllegalArgumentException, and NullPointerException, or is it acceptable to catch a general Exception here since multiple types could be thrown?

HamzaElzarw-2022: I found that MalformedURLException doesn’t provide a constructor that takes an exception as an argument. So I used initCause() and casted the result back to MalformedURLException like this:

`throw (MalformedURLException) new MalformedURLException(e.getMessage()).initCause(e);`

Could you please confirm if this implementation is acceptable?

sebastian-nagel:
Hi @HamzaElzarw-2022, thanks and sorry for the delayed response.

I'd be fine with your implementation. It's a good starting point.

I've started to implement a unit test - for now it's a very early draft: #536. TODOs from the failing tests:
- address the bug with relative URLs (see #522)
- eventually allow the base URL to be null. This is allowed in the URL constructor if "spec" is an absolute URL:
  ```java
  new URL(null, "https://example.org/")
  ```
  Otherwise this should be documented. Also: throwing a NPE seems to be better in this case.
- there is an error if the "spec" starts with a colon:
  ```jshell
  jshell> new URL(new URL("https://example.org/foo/"), ":23")
  $7 ==> https://example.org/foo/:23

  jshell> new URI("https://example.org/foo/").resolve(":23")
  |  Exception java.lang.IllegalArgumentException: Expected scheme name at index 0: :23
  ...
  ```
- eventually implement a similar method for URIs as well
  
@HamzaElzarw-2022, let me know whether you want to continue to work on this PR. Otherwise, I'm going to merge it. We can address the open points later. And of course, feel free to pick from #536. Thanks!

HamzaElzarw-2022:
Hello @sebastian-nagel, I will happily continue working on the PR.

HamzaElzarw-2022:
Hello @sebastian-nagel, sorry for the very late follow-up.

I’ve addressed all the points you mentioned earlier, and all tests are now passing. Please review it when you have a moment.

To fix the bug with relative URLs described in NUTCH-797, I prepended the last segment to the spec to reverse the effect of URI.resolve().

src/main/java/crawlercommons/utils/URLUtils.java
```diff
@@ -0,0 +1,44 @@
...
+                String basePath = base.getPath();
+                if (basePath != null && !basePath.isEmpty()) {
+                    int lastSlash = basePath.lastIndexOf('/');
+                    if (lastSlash > -1 && lastSlash < (basePath.length() - 1)) {
+                        resolvedSpec = basePath.substring(lastSlash + 1) + spec;
+                    }
+                }
+            }
+
+            return base.toURI().resolve(resolvedSpec).normalize().toURL();
```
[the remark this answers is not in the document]
HamzaElzarw-2022: I added `normalize()` in the first place to normalize in case it prepended `./` for when spec starts with a `:` but as you mentioned `URL::resolve` normalize relative paths any way. so in this case I think we should remove it as normalizing the base shouldn’t be part of this method’s responsibilities.

HamzaElzarw-2022:
Great! After this is merged, I’ll open another PR to replace usages of URI.resolve() with this utility.

sebastian-nagel:
Thanks, @HamzaElzarw-2022 and @rzo1!

PR #530, opened 2025-07-16
Add matchedWildcard flag to BaseRobotRules

CoGiang:
This PR adds a getter/setter for the isMatchedWildcard flag to the BaseRobotRules class. This allows the client be able to tell whether the passed in user agent was explicitly matched, or it defaulted to the wildcard user agent.

This is useful for reporting, where we can distinguish between which hosts matched a specific rule vs a wildcard.

### DIFF
--- a/src/main/java/crawlercommons/robots/BaseRobotRules.java
+++ b/src/main/java/crawlercommons/robots/BaseRobotRules.java
@@ -44,7 +44,7 @@ public abstract class BaseRobotRules implements Serializable {
 
     private long _crawlDelay = UNSET_CRAWL_DELAY;
     private boolean _deferVisits = false;
-    protected boolean _matchedWildcard = false;
+    private boolean _matchedWildcard = false;
     private LinkedHashSet<String> _sitemaps;
 
     public BaseRobotRules() {
@@ -86,7 +86,7 @@ public abstract class BaseRobotRules implements Serializable {
 
     /**
      * Returns whether the wildcard user agent (*) was matched.
-     *
+     * 
      * @return true if the wildcard user agent was matched; false otherwise.
      */
 
@@ -96,8 +96,9 @@ public abstract class BaseRobotRules implements Serializable {
 
     /**
      * Sets whether the wildcard user agent (*) was matched.
-     *
-     * @param matchedWildcard true if the wildcard user agent was matched; false otherwise.
+     * 
+     * @param matchedWildcard
+     *            true if the wildcard user agent was matched; false otherwise.
      */
     public void setMatchedWildcard(boolean matchedWildcard) {
         this._matchedWildcard = matchedWildcard;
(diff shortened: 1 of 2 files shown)


CoGiang:
Hi @sebastian-nagel, could you review the PR when you have some time, thanks!

src/main/java/crawlercommons/robots/BaseRobotRules.java
```diff
@@ -44,6 +44,7 @@ public abstract class BaseRobotRules implements Serializable {
 
     private long _crawlDelay = UNSET_CRAWL_DELAY;
     private boolean _deferVisits = false;
+    protected boolean _matchedWildcard = false;
```
sebastian-nagel, line 47: I thought longer where to best store the variable?
1. SimpleRobotRules because also the allow/disallow rules are placed there.
2. (here) BaseRobotRules:   - we hold the Crawl-delay and Sitemaps here: the former is specific to one user-agent, the later is global for all user-agents
3. SimpleRobotRulesParser:
   - what was matched requires to actually parse a robots.txt
   - also: the variable is (as of now) not part of the hashCode, equals, toString methods of the rules
   - keeping it in SimpleRobotRulesParser is somewhat ugly because the parser is stateless:
     - there are variables configuring the parser behavior (eg. `_exactUserAgentMatching`)
     - but only one preserves state from the last parse `_numWarningsDuringLastParse`. It's used purely for logging / debugging. But this feature is also about "reporting".

Another point: the variable could be also a private one. Or is there a reason it's `protected`?

sebastian-nagel (changes requested):
Hi @CoGiang,

thanks for the contribution!

It is definitely a useful feature and the implementation looks good.

Are there any comments or definite objections where to place the variable whether the wildcard user-agent was matched? The decision feels not easy, also because our robots.txt parser is not 100% consequent here (for historic reasons). So, no strict complaints against your solution, @CoGiang.

Two marginal points:
- I'd vote to make the variable private (but this is no blocker!)
- @CoGiang, could you run `mvn java-formatter:format` and update the PR with the format changes? Thanks!

src/main/java/crawlercommons/robots/BaseRobotRules.java
CoGiang: Thanks for the comments @sebastian-nagel!

Yeah, it did seem awkward to put it in SimpleRobotRulesParser because of the points you mentioned. The matchedWildcard flag seems more like a rule level than a parse thing. Especially, if you want to use different user agent combinations with a single parser.
I put it in BaseRobotRules since BaseRobotRules also had the crawl-delay, sitemaps, and isDeferVisit fields. But I don't really have a strong opinion whether to store it in SimpleRobotRules or BaseRobotRules.

I thought I already made the field private but I forgot to push those changes. No reason for it to be `protected`.

sebastian-nagel: > Especially, if you want to use different user agent combinations with a single parser.

Yes, definitely.

PR #531, opened 2025-07-19
Replace deprecated url basic normalizer

HamzaElzarw-2022:
**Description**:
Replaces all deprecated usages of `new URL(...)` constructors in `BasicURLNormalizer.java` which is the only remaining implementation class with such deprecated constructors.

**Details**:
Directly replacing new URL(...) with URI.toURL() was not feasible in all cases due to behavioral differences between URL and URI. The reasoning and applied solutions are outlined below:

1. Handling `new URL(String url)` in `parseStringToURL` method
The method `parseStringToURL`, used at the beginning of `filter` method to extract URL components, used new URL(String) to deconstruct the input string. Replacing this with `URI` was not viable, as `URI` is stricter and would throw exceptions for cases that were meant to be handled later in the `filter` method (e.g., URLs containing spaces or non-ASCII characters).
To address this, a new method `deconstructURL` was introduced, which uses a `urlDeconstructionPattern` regex to extract components without instantiating a URL prematurely.

2. Handling `new URL(protocol, host, port, file)` constructor
Replacing this form with its closest URI substitute  `new URI(schema, userinfo, host, port, path, query, fragment)` led to unwanted percent encoding behavior. For instance:

```
test case: org.opentest4j.AssertionFailedError: normalizing: http://foo.com/%66oo.htm%C0 ==> 
Expected :http://foo.com/foo.htm%C0 
Actual :http://foo.com/foo.htm%25C0
```

To avoid this, the URL is reconstructed as a string from its components after normalization, and then converted using new URI(String).toURL() to preserve encoding.

**Other refactoring details**:
- The changed flag is now only used after the final URL is constructed, since a URL instance will be created regardless.
- Temporary variables like file2 and newHost, previously used just to track changes, were removed for clarity.
- The check for whether the port equals the default port was moved after the initial URL construction as it calls `url.getDefaultPort()`.

I tried to justify the changes as possible hope it makes sense :) 
I’d be happy to hear any feedback and will gladly update the PR to simplify or reduce the changes as possible.

### DIFF
--- a/src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
+++ b/src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
@@ -22,9 +22,20 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.UnsupportedEncodingException;
-import java.net.*;
-import java.sql.SQLOutput;
-import java.util.*;
+import java.net.IDN;
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.net.URLDecoder;
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -182,6 +193,7 @@ public class BasicURLNormalizer extends URLFilter {
 
         String[] urlParts = deconstructUrl(urlString);
 
+        // fallback to http:// if couldn't deconstruct or no protocol is specified
         if (urlParts == null || urlParts[0] == null) {
             return filter("http://" + urlString);
         }
--- a/src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
+++ b/src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
@@ -22,11 +22,8 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.UnsupportedEncodingException;
-import java.net.IDN;
-import java.net.MalformedURLException;
-import java.net.URISyntaxException;
-import java.net.URL;
-import java.net.URLDecoder;
+import java.net.*;
+import java.sql.SQLOutput;
 import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


HamzaElzarw-2022:
I forgot to mention that there's one failing test:

```
org.opentest4j.AssertionFailedError: normalizing: file:///foo/bar.txt ==> 
Expected : file:///foo/bar.txt
Actual   : file:/foo/bar.txt
```

This seems to be caused by normalization behavior within the URI constructor itself as I explicitly add `://` to the url string before passing it to the constructor and still URI removes it again. I'm not exactly sure how to work around this, so any guidance would be appreciated.

rzo1:
`file:///foo/bar.txt` is the standard and correct format of a file URI, so I would expect it in the result (according to https://datatracker.ietf.org/doc/html/rfc3986)

rzo1:
Yep, I can confirm

```
[ERROR] Failures: 
[ERROR]   EffectiveTldFinderTest.testMatchAllSuffixes:331 Domain under public suffix is expected to be same as host ==> expected: <foo.bar.uberspace.de> but was: <uberspace.de>
[ERROR]   EffectiveTldFinderTest.testMatchAllSuffixes:337 Host name is a wildcard public suffix ==> expected: <null> but was: <uberspace.de>
```
from master.

rzo1:
@HamzaElzarw-2022 Think you can rebase now, so we get a clean view of the build

HamzaElzarw-2022:
Hi @rzo1,
Any updates on this PR? There's only one failing case remaining, and it seems directly related to the behavior of `URI`. Do you think it's worth implementing a workaround to fix it, or is it something we can ignore? I can update the test if needed, just let me know what you think.

sebastian-nagel:
Regarding the failing unit test:
- After digging into the code and its history: it was originally added to ensure that a malformed input does not cause a NPE, see [here](https://github.com/apache/nutch/blob/8416da8a16dba003dbc8d0bbc3c5b0e997301780/src/plugin/urlnormalizer-basic/src/test/org/apache/nutch/net/urlnormalizer/basic/TestBasicURLNormalizer.java#L178).
- For sure, you'll find a `<a href="http:">...</a>` somewhere in the web, and a NPE in the normalizer should not cause a crawler job to fail.

For now, we could just comment out the corresponding test (in weirdToNormalizedUrls.csv) and open a separate issue to address it later: test for the NPE. Since the URL is not crawlable anyway, the exact output does not matter. A `null` would be the best return value.

src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
```diff
@@ -244,23 +261,25 @@ public String filter(String urlString) {
             // check for unnecessary use of "/../", "/./", and "//"
             try {
                 if (changed) {
-                    url = new URL(protocol, host, port, file);
+                    String tempUrl = protocol + "://" + (host == null ? "" : host) + (port == -1 ? "" : ":" + port) + file;
```
sebastian-nagel, line 264: This seems overtly complex. We could just return the concatenated string. Except we want explicitly verify that the normalized URL is both a valid URI and URL.

sebastian-nagel: ... or use the URI internally for now. In `getFileWithNormalizedPath(url)` we convert again from URL to URI.

src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
```diff
@@ -277,13 +296,13 @@ public String filter(String urlString) {
...
         URL url = null;
         try {
-            url = new URL(urlString);
-        } catch (MalformedURLException e) {
+            url = new URI(urlString).toURL();
+        } catch (MalformedURLException | URISyntaxException | IllegalArgumentException e) {
             if (!hasSchemePattern.matcher(urlString).find()) {
                 // no protocol/scheme : try to prefix http://
                 try {
-                    url = new URL("http://" + urlString);
```
sebastian-nagel, line 285: Ok, here might be the origin for the failing unit test: "http://" is added in front of "http:".

sebastian-nagel (changes requested):
@HamzaElzarw-2022, eventually the code can be simplified if the URI class is directly used without converting from and to URL. As far as I can see, all methods which take a URL object as parameter or returning one are private. It would be safe to change them.

For the failing unit test: see comments.

HamzaElzarw-2022:
Looks clean @sebastian-nagel, thank you!

sebastian-nagel:
Thanks, @HamzaElzarw-2022 and @rzo1!

PR #544, opened 2025-11-04
[Domains] Update unit tests after change in public suffix list

sebastian-nagel:
- a change in the public suffix list (publicsuffix/list@30f4f42) causes a unit test failure
- this PR updates unit test: replace `*.bd` by `*.ck`

While the PSL (at least, the ICANN section) wasn't subject to frequent changes over several years, this was the second change this year causing an issue (after #532). Eventually, we should change the way how EffectiveTLFinder is tested.


rzo1 (approved):

sebastian-nagel:
Thanks, @rzo1!

Issue #551, opened 2025-11-29
PublicSuffix List introduced ᬩᬮᬶ.id which breaks tests


rzo1:
```
// xn--9tfky.id (<bali>.id, Und-Bali)
ᬩᬮᬶ.id
```

for example:

```
Error:    PaidLevelDomainTest.testStandardDomains:56 » IllegalArgument java.text.ParseException: An unassigned code point was found in the input ᬩᬮᬶ
```

aecio:
I'm a bit late for the conversation, but I agree that 13.2M added by icu4j would not be too big for crawler projects, especially if crawler commons also had a class for charset identification (#171). But it seems like #551 already fully fixes the problem.

rzo1:
@aecio Thanks for your feedback. I wanted to restore our CI build asap :) - #551 fixed the initial issue, yes. Maybe we can consider icu4j, if something similar occurs in the near future.

PR #552, opened 2025-12-01
#551 – Support IDNA2008 Unicode domains by using ALLOW_UNASSIGNED in IDN methods within TldFinder / Normalizer.

rzo1:
We actually didn't use the constant at all.


rzo1:
Thanks for the review @sebastian-nagel

PR #569, opened 2026-02-18
Support LLM-Policy entry in robots.txt

lewismc:
Proposed patch for https://github.com/crawler-commons/crawler-commons/issues/564

This PR adds a pluggable extension API to the crawler-commons robots.txt parser, enabling opt-in support for non-standard directives such as LLM-Policy, Clean-Param, Content-Signals, and others. The design follows the same philosophy as the existing sitemap Extension API but is adapted to the simpler, line-oriented nature of robots.txt.

Backwards compatibility:
* No breaking changes. By default no extensions are enabled -- behavior is identical to before.
* The `_extensions` map in `BaseRobotRules` stays `null` until a value is added, so there is _zero memory overhead_ for crawlers that don't use extensions.
Existing `getCrawlDelay()` / `getSitemaps()` methods are unchanged; they remain part of the core contract.

Example usage:
```java
SimpleRobotRulesParser parser = new SimpleRobotRulesParser();

// Enable a single extension
parser.enableExtension(RobotsExtension.LLM_POLICY);

// Enable a specific set of extensions
parser.enableExtensions(EnumSet.of(
    RobotsExtension.LLM_POLICY,
    RobotsExtension.CONTENT_SIGNALS,
    RobotsExtension.REQUEST_RATE
));

// Or enable all known extensions at once
parser.enableAllExtensions();

// Inspect which extensions are currently enabled
Set<RobotsExtension> enabled = parser.getEnabledExtensions();

// Parse as usual -- extension data is collected automatically
SimpleRobotRules rules = parser.parseContent(url, content, contentType, agentNames);

// Access data for a specific extension
RobotsExtensionData llmPolicy = rules.getExtensionData(RobotsExtension.LLM_POLICY);
if (llmPolicy != null) {
    List<String> values = llmPolicy.getValues(); // e.g. ["/llms.txt"]
    Map<String, String[]> map = llmPolicy.asMap(); // {"llm-policy": ["/llms.txt"]}
}

// Access all collected extension data at once
Map<RobotsExtension, RobotsExtensionData> allExtensions = rules.getExtensions();
```

### DIFF
--- a/src/main/java/crawlercommons/robots/RobotsExtension.java
+++ b/src/main/java/crawlercommons/robots/RobotsExtension.java
@@ -1,0 +1,144 @@
+/**
+ * Copyright 2016 Crawler-Commons
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package crawlercommons.robots;
+
+/**
+ * Known robots.txt extension directives that can be optionally enabled in
+ * {@link SimpleRobotRulesParser}.
+ *
+ * <p>
+ * Each constant carries:
+ * </p>
+ * <ul>
+ * <li>{@link #getDirectiveName()} &ndash; the canonical lower-case directive
+ * name as it appears in a robots.txt file (e.g. {@code "llm-policy"})</li>
+ * <li>{@link #isPerGroup()} &ndash; whether the directive is scoped to a
+ * specific user-agent group ({@code true}) or applies globally to the entire
+ * robots.txt file ({@code false})</li>
+ * <li>{@link #getAliases()} &ndash; optional alternative spellings that should
+ * also be recognized</li>
+ * </ul>
+ *
+ * <p>
+ * Extensions are opt-in: they are only parsed and stored when explicitly enabled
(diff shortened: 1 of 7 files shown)


lewismc:
@rzo1 I like the co-pilot review. So I am clear, is the policy to adopt suggestions or for them to complement human review? Thank you.

rzo1:
> @rzo1 I like the co-pilot review. So I am clear, is the policy to adopt suggestions or for them to complement human review? Thank you.

I am currently short of time, so I thought it might be valueable to trigger an AI review ;-)
I think it is actually complemental to a human review but might catch things people tend to oversee.

So if you think the suggestion make sense, feel free to integrate (imho)

src/main/java/crawlercommons/robots/RobotsExtension.java
```diff
@@ -0,0 +1,144 @@
+/**
+ * Copyright 2016 Crawler-Commons
```
rzo1: Can be 2026 since it is new code; also for all other new files

rzo1 (approved):
I only have a minor nit regarding copyright year. Otherwise, lgtm. thanks.

lewismc:
Im happy to keep this open as long as required for more peer review. Thank you.

src/main/java/crawlercommons/robots/RobotsExtensionData.java
```diff
@@ -0,0 +1,55 @@
+/**
+ * Copyright 2016 Crawler-Commons
```
sebastian-nagel: The year indicates "2016" while in RobotsExtension.java it's "2026".

sebastian-nagel (approved):
Thanks, @lewismc! Great contribution.

Sorry, it took long to review and test this PR. (Test code [here](https://github.com/sebastian-nagel/sitemap-performance-test/pull/1)).

src/main/java/crawlercommons/robots/RobotsExtensionData.java
lewismc: Good catch @sebastian-nagel

lewismc:
Thanks for the review.

rzo1:
thanks for the work @lewismc

PR #584, opened 2026-06-30
Add lazy CrawlerURL and URI/CrawlerURL API variants (#556)

### LINKED ISSUE #556: Provide variant API methods taking URL parameters as URI or String
Crawler-Commons uses a mixture of [java.net.URL](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/net/URL.html)s or Strings as method parameter, e.g.,
- [SimpleRobotRulesParser.parseContent](https://crawler-commons.github.io/crawler-commons/1.5/crawlercommons/robots/SimpleRobotRulesParser.html#parseContent(java.lang.String,byte[],java.lang.String,java.util.Collection)) takes a String, while
- [SiteMapParser.parseSiteMap](https://crawler-commons.github.io/crawler-commons/1.5/crawlercommons/sitemaps/SiteMapParser.html#parseSiteMap(byte%5B%5D,java.net.URL)) expects a URL.

Due to the deprecation of java.net.URL constructors, also [URI](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/net/URI.html)s were introduced into the crawler-commons codebase, see [#522](https://github.com/crawler-commons/crawler-commons/issues/522).

Naturally, a crawler has to deal with URLs. Different Java-based crawler implementations use different classes to represent URLs:
- [Nutch](https://nutch.apache.org/) and [StormCrawler](https://stormcrawler.apache.org/) rely mostly on the older URL class (since Java 1.0) and not the newer URI class (since 1.4)
- [Heritrix](https://github.com/internetarchive/heritrix3/) uses mostly the URI class
  - web archiving crawlers prefer URI because the WARC standard requires URIs
  - however, if used libraries require it, URL objects are used
- [BUbiNG](https://github.com/LAW-Unimi/BUbiNG) uses URIs but also has its own wrapper class [BURL](https://law.di.unimi.it/software/bubing-docs/index.html?it/unimi/di/law/bubing/util/BURL.html).

There are subtle API differences between URL and URI, and also some issues with the URL class. See [#522](https://github.com/crawler-commons/crawler-commons/issues/522) and the URL methods listed in [forbidden-apis-signatures.txt](https://github.com/crawler-commons/crawler-commons/blob/master/src/test/resources/forbidden-apis-signatures.txt).

These points are good arguments for crawler-commons to **additionally** provide methods that take URIs and/or strings as arguments, so that web crawler implementations are not forced to convert from URIs to URLs when using crawler-commons. Internally, we may use URL or URI, whichever is more efficient.

Comments on this proposal are welcome!

rzo1:
## Summary

Resolves #556. The crawler-commons API inconsistently accepts `String` or
`java.net.URL`, never `java.net.URI`, forcing downstream crawlers (Nutch,
StormCrawler, Heritrix, BUbiNG) into wasteful conversions between URL
representations. Following the discussion in #556, this introduces a custom
URL class that lazily converts between representations and caches the results,
plus backward-compatible overloads across the public API.

### DIFF
--- a/src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
+++ b/src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
@@ -275,7 +275,7 @@ public class BasicURLNormalizer extends URLFilter {
                     String tempUrl = protocol + "://" + (host == null ? "" : host) + (port == -1 ? "" : ":" + port) + file;
                     file2 = getFileWithNormalizedPath(new URI(tempUrl));
                 } else {
-                    file2 = getFileWithNormalizedPath(new URI(urlString));
+                    file2 = getFileWithNormalizedPath(crawlerUrl.toJavaURI());
                 }
                 if (!file.equals(file2)) {
                     changed = true;
--- a/src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
+++ b/src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
@@ -44,6 +44,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import crawlercommons.filters.URLFilter;
+import crawlercommons.url.CrawlerURL;
 
 /**
  * Converts URLs to a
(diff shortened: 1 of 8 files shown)


src/test/java/crawlercommons/url/CrawlerURLTest.java
```diff
@@ -0,0 +1,185 @@
...
+        CrawlerURL a = CrawlerURL.of(SAMPLE);
+        CrawlerURL b = CrawlerURL.of(SAMPLE);
+        CrawlerURL c = CrawlerURL.of("http://other.example.org/");
+
+        assertEquals(a, b);
+        assertEquals(a.hashCode(), b.hashCode());
+        assertNotEquals(a, c);
+        assertEquals(SAMPLE, a.toString());
+        assertEquals(a.toStringURL(), a.toString());
+        assertNotEquals(a, null);
```
sebastian-nagel: Could swap args: expected, actual value.

sebastian-nagel: Or use `assertNotNull` here.

Swap args one line below.

src/main/java/crawlercommons/filters/basic/BasicURLNormalizer.java
sebastian-nagel: Although it makes no difference in the result, could call `crawlerUrl.toJavaURI()` here.

sebastian-nagel: It might even make a tiny little difference, in case `parseStringToURL` resp. `parseStringToCrawlerURL` adds a schema (`http://`). That's rather a bug in the existing implementation.

sebastian-nagel (approved):
Thanks, @rzo1!

Excellent work. Code looks good, documentation also.

I've verified on a 10 million sample of URLs that BasicURLNormalizer did not change the behavior.

rzo1:
Adressed comments. Thanks for the review

PR #586, opened 2026-06-30
Port WHATWG URL tests from Web Platform Tests (#535)

### LINKED ISSUE #535: Port URL unit tests from Web Platform Tests
The web-platform-tests project (<https://web-platform-tests.org/>) includes a URL test suite verifying compatibility with the whatwg URL spec (<https://url.spec.whatwg.org/>).

It would be nice to integrate theses tests into crawler-commons, testing the new URLUtils class (#525) in combination with BasicURLNormalizer.

- test data: <https://github.com/web-platform-tests/wpt/blob/master/url/resources/urltestdata.json>
- see also: <https://github.com/whatwg/url>

rzo1:
Closes #535.

Validates the crawler-commons URL pipeline — `URLUtils.resolve` (#525) feeding `BasicURLNormalizer` — against the WHATWG URL Standard by running the [Web Platform Tests](https://github.com/web-platform-tests/wpt) `url/resources/urltestdata.json` corpus.

### What's included

- **Vendored corpus** — `urltestdata.json` pinned to wpt commit `6778f81bd5408ae4b8ca9e800ffaca59e65688d9`, with a `wpt-urltestdata-LICENSE.txt` carrying the BSD-3-Clause text and source/commit attribution.
- **`SimpleJsonArrayParser`** — a small, dependency-free reader sufficient for this corpus (a JSON array of comment strings or flat objects; values are string/boolean/null), with full escape handling including `\uXXXX`. Comes with its own unit test. No new build/runtime dependency is added.
- **`WptUrlNormalizationTest`** — a parameterized test that filters the corpus to the in-scope `http`/`https` subset (skipping opaque-path bases and non-special schemes), resolves each `input` against its `base` via `URLUtils.resolve`, normalizes the result, and asserts it equals the WHATWG `href`. Entries marked `failure: true` are asserted to be rejected (null / `MalformedURLException`).

### Results

- **231** in-scope cases pass.
- **201** cases where the Java `URI` + crawler-commons pipeline structurally diverges from WHATWG are recorded in a documented `EXCLUSIONS` set, each with an inline reason and tagged:
  - **`WHATWG-divergence` (197)** — userinfo stripping, fragment dropping, tab/newline & C0 trimming, backslash-as-path-separator, IPv4/IPv6 host canonicalization, percent-encode sets, IDNA/UTS-46, etc. These reflect the gap between `java.net.URI` semantics and the WHATWG spec.
  - **`cc-gap` (4)** — likely crawler-commons normalization bugs worth a follow-up:
    - `http://example.com/foo/%2e./%2e%2e/.%2e/%2e.bar` — dot-segment removal corrupts the authority (yields host `example.combar`)
    - `http://example.com////../..`, `…/foo/bar//../..`, `…/foo/bar//..` — trailing empty-segment collapse during dot removal


src/test/java/crawlercommons/utils/WptUrlNormalizationTest.java
```diff
@@ -0,0 +1,419 @@
...
+        EXCLUSIONS.add("http:\\\\a\\b:c\\d@foo.com\\ || http://example.org/foo/bar"); // WHATWG-divergence: backslash treated as path separator for special schemes; Java URI rejects
+        EXCLUSIONS.add("http://a:b@c\\ || "); // WHATWG-divergence: backslash treated as path separator for special schemes; Java URI rejects
+        EXCLUSIONS.add("http://foo/path;a??e#f#g || http://example.org/foo/bar"); // WHATWG-divergence: WHATWG percent-encodes forbidden chars in path/query/fragment; Java URI rejects
+        EXCLUSIONS.add("http://foo/abcd#foo?bar || http://example.org/foo/bar"); // WHATWG-divergence: crawler-commons drops the fragment component
+        EXCLUSIONS.add("[61:24:74]:98 || http://example.org/foo/bar"); // WHATWG-divergence: IPv6 literal parsing/compression; Java URI rejects
+        EXCLUSIONS.add("http:[61:27]/:foo || http://example.org/foo/bar"); // WHATWG-divergence: IPv6 literal canonicalization differs
+        EXCLUSIONS.add("http://[2001::1] || http://example.org/foo/bar"); // WHATWG-divergence: IPv6 literal parsing/compression; Java URI rejects
```
sebastian-nagel, line 110:
```
jshell> new URI("http://[2001::1]").toURL()
$1 ==> http://[2001::1]
```

That's actually not a divergence but a bug in BasicURLNormalizer. Reported as #587.

sebastian-nagel (approved):
Thanks, @rzo1! Great! contribution!

Apart from #587, the list of exclusions might a need careful review which of the "divergences" are rather bugs or nice-to-have features. Canonicalization of IP addresses is one candidate because it's a potential source of duplicates. But that's not necessarily the task now. It might even need several passes.

rzo1:
Yes. I think we can treat it as follow-up. Thanks for the review!

PR #590, opened 2026-07-24
Fix per-group robots.txt extension values leaking from wildcard group

jnioche:
## Problem

When a specific `User-agent` group supersedes an earlier wildcard (`*`) group, `handleUserAgent` clears the wildcard's allow/disallow rules and crawl-delay — but it did **not** clear per-group extension data (e.g. `Clean-param`, `Request-rate`, `No-index`, `Visit-time`, `Comment`, `Robot-version`) collected under the wildcard group. Those values leaked into the matched agent's rules.

### Reproduction

```
User-agent: *
Request-rate: 1/10m
Disallow: /wild/

User-agent: mybot
Disallow: /bot/
```

Parsing for `mybot` (with `REQUEST_RATE` enabled) correctly drops `Disallow: /wild/`, but `getExtensionData(REQUEST_RATE)` still returns `1/10m` from the wildcard group.

## Fix

Clear per-group extension data alongside the rules and crawl-delay when a matching specific user-agent group replaces the wildcard group (both the exact-matching and legacy prefix-matching paths). Global / file-wide extensions (`LLM-Policy`, `Host`, `Content-Signals`) are intentionally unaffected.

## Tests

Added `testExtensionPerGroupNotLeakedFromWildcard`. Full `SimpleRobotRulesParserTest` suite passes (233 tests).

### DIFF
--- a/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
+++ b/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
@@ -296,6 +296,21 @@ public class SimpleRobotRulesParser extends BaseRobotsParser {
             _curRules.clearRules();
         }
 
+        /**
+         * Clear the per-group extension data collected so far in the current
+         * rules. Called when a wildcard group is superseded by a matching
+         * specific user-agent group, so that per-group extension values (e.g.
+         * Clean-param) collected under the wildcard do not leak into the rules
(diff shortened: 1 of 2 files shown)


sebastian-nagel (approved):
Thanks, @jnioche. Good catch!

While reviewing, I've found that there is no unit test yet testing group merging in combination with extensions. One is added in #594. This PR works perfectly for group merging.

PR #591, opened 2026-07-24
Do not add unescaped Allow path when path normalization fails

jnioche:
## Problem

`handleAllow()` calls `normalizePathDirective()` inside a `try`/`catch`, but adds the rule **after** the catch block:

```java
try {
    path = normalizePathDirective(path);
} catch (Exception e) {
    reportWarning(state, "Error parsing robots rules - can't decode path: {}", path);
}
if (path.length() == 0) { ... } else {
    state.addRule(path, true);   // still runs on exception, with the raw path
}
```

So when normalization/decoding fails, the parser logs a warning **and then still installs an allow rule using the raw, un-normalized path**. The sibling `handleDisallow()` keeps `addRule()` inside the `try`, so on failure it drops the rule instead.

## Fix

Move the rule creation inside the `try` block so `handleAllow()` behaves identically to `handleDisallow()`: a normalization failure results in no rule being added rather than a rule with an unescaped path.

## Tests

`SimpleRobotRulesParserTest` suite passes (232 tests).

### DIFF
--- a/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
+++ b/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
@@ -1095,18 +1095,17 @@ public class SimpleRobotRulesParser extends BaseRobotsParser {
 
         try {
             path = normalizePathDirective(path);
+            if (path.length() == 0) {
+                /*
+                 * Allow: <nothing> => allow all.
+                 *
+                 * See handleDisallow(...): We ignore the empty allow statement.
+                 */
+            } else {
+                state.addRule(path, true);
+            }
         } catch (Exception e) {
             reportWarning(state, "Error parsing robots rules - can't decode path: {}", path);
-        }
-
-        if (path.length() == 0) {
-            /*
-             * Allow: <nothing> => allow all.
-             * 
-             * See handleDisallow(...): We ignore the empty allow statement.
-             */
-        } else {
-            state.addRule(path, true);
         }
     }
 


sebastian-nagel (approved):
Good catch!

The question is, why there is a try-catch block at all? No thrown exception is declared. I also cannot see anything which would throw a RuntimeException. Also: grepped log files of robots.txt parsing of 10 million hosts. The error message wasn't there. Maybe this is a relict of an earlier version of the `normalizePathDirective` method.

jnioche:
@sebastian-nagel you are right, the catch (Exception e) blocks — at line 1153 (handleDisallow) and line 1175 (handleAllow), both reporting "Error parsing robots rules - can't decode path" — are no longer needed. Nothing inside their try blocks can throw anymore. They are dead code, kept around after the operation they were protecting was replaced.

The try bodies do only:

- normalizePathDirective(path) → SimpleRobotRules.escapePath(path.trim(), null) → BasicURLNormalizer.escapePath(BasicURLNormalizer.unescapePath(path))
- (in handleDisallow only) state.addRule(path, false)

None of these throw:
- unescapePath (BasicURLNormalizer.java:526) scans with the regex %([0-9A-Fa-f]{2}), so Integer.valueOf(group, 16) only ever sees two hex digits (≤ 0xFF) — it can't throw, and all StringBuilder indices are in range.
- escapePath (BasicURLNormalizer.java:573) is pure byte manipulation and is deliberately tolerant — e.g. a stray % is emitted as %25 rather than raising an error.
- addRule (SimpleRobotRules.java:161) just appends to a list.

What needed the catch historically

Originally the body was a call to URLDecoder.decode(path, "UTF-8"). In the very first version (cd55b6b, "Preliminary versions of robots.txt processing code"), handleDisallow/handleAllow looked like:

try {
    path = URLDecoder.decode(path, "UTF-8");
    ...
} catch (Exception e) {
    reportWarning("Error parsing robots rules - can't decode path: " + path, ...);
}

URLDecoder.decode throws:
- IllegalArgumentException (unchecked) on a malformed percent-escape in the robots.txt path — e.g. a % not followed by two hex digits. This is the real reason the catch existed: robots.txt files in the wild contain broken escapes, and this turned a crash into a logged warning.
- (Historically also UnsupportedEncodingException, though never with a literal "UTF-8".)

The URLDecoder.decode call was removed in commit 9559134 ("Path analyse bug with url-decode if allow/disallow path contains escaped wild-card characters", refined in 5d036a1), which replaced ad-hoc URL-decoding with the percent-encoding-aware normalizePathDirective/escapePath path. That new code is non-throwing, but the surrounding try/catch (and its "can't decode path" wording) was carried over unchanged and has been vestigial ever since.

(not me - Claude doing the detective work).

Will open a new PR to remove the dead catch blocks

PR #592, opened 2026-07-24
Avoid NullPointerException in getNumWarnings() before first parse

jnioche:
## Problem

`getNumWarnings()` returns the value of a `ThreadLocal<Integer>` that is only assigned inside `parseContent()`:

```java
private ThreadLocal<Integer> _numWarningsDuringLastParse = new ThreadLocal<>();
...
public int getNumWarnings() {
    return _numWarningsDuringLastParse.get();   // auto-unboxing null -> NPE
}
```

If `getNumWarnings()` is called on a parser (or thread) that has not yet parsed a robots.txt file, `get()` returns `null` and auto-unboxing to `int` throws a `NullPointerException`.

## Fix

Initialize the `ThreadLocal` with `ThreadLocal.withInitial(() -> 0)` so it returns `0` before the first parse instead of `null`.

## Tests

Added `testGetNumWarningsBeforeParse`. Full `SimpleRobotRulesParserTest` suite passes (233 tests).


sebastian-nagel (approved):
+1

PR #593, opened 2026-07-24
Use a prefix trie in tokenize() instead of scanning every directive prefix

jnioche:
## Motivation

`tokenize()` scanned the **entire** `DIRECTIVE_PREFIX` map (~40 entries) for every robots.txt line, calling `startsWith()` on each in non-deterministic `HashMap` iteration order:

```java
for (String prefix : DIRECTIVE_PREFIX.keySet()) {
    if (lowerLine.startsWith(prefix)) { ... }
}
```

That is O(number of prefixes) per line and order-dependent.

## Change

Build a **prefix trie** once (in the existing static block) from the same `DIRECTIVE_PREFIX` map — which stays as the readable source of truth, including all the typo aliases. `tokenize()` now walks the line character by character through the trie, running the separator/data check only at nodes where a known directive prefix ends, and stops as soon as no child matches. This is O(matched-prefix length) per line and deterministic.

The per-prefix separator matching (the colon/blank delimiter logic and the `isPrefix` ACAP handling) is extracted **unchanged** into a `matchDirective()` helper, so each line is classified exactly as before.

## Verification

- Full test suite passes (518 tests).
- **Behavioural equivalence**: compared the new trie against a faithful reimplementation of the previous algorithm over **2,853,278 distinct real-world robots.txt lines** — **0 differences**.
- **Performance**: a micro-benchmark of both algorithms over the same corpus shows a consistent **~1.2× speedup** on the tokenize step (the remainder is the shared regex work both versions perform).

No behavioural change is intended or observed; this is a performance/clarity refactor.

### DIFF
--- a/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
+++ b/src/main/java/crawlercommons/robots/SimpleRobotRulesParser.java
@@ -394,6 +394,20 @@ public class SimpleRobotRulesParser extends BaseRobotsParser {
 
     private static Map<String, RobotDirective> DIRECTIVE_PREFIX = new HashMap<String, RobotDirective>();
 
+    /**
+     * Node of the directive-prefix trie used by {@link #tokenize(String)} to
+     * look up the directive at the start of a robots.txt line without scanning
+     * every known prefix. Each node represents the prefix spelled out by the
+     * path from the root; {@link #directive} is non-{@code null} if a known
+     * directive prefix ends at this node.
+     */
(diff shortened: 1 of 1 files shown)


sebastian-nagel (approved):
Excellent and lean improvement! Thanks, @jnioche.

Tested with a robots.txt test corpus with all available extensions enabled. Performance has improved by 5-10% overall, including parsing of all directives (not only extensions).

PR #594, opened 2026-07-26
[Robots] Add unit tests to verify that group merging and extension data

sebastian-nagel:
The robots.txt parser tests lack unit tests which verify that
- extension data from multiple robots.txt rule groups is merged,
- the Crawl-delay is properly assigned to a single user-agent, even if there are multiple rule groups for this user-agent,
- and the extensions do not interfere with the merging of allow/disallow rules.

Merging of groups is described in [RFC 9309, section 2.2.1](https://www.rfc-editor.org/rfc/rfc9309.html#section-2.2.1):
> If there is more than one group matching the user-agent, the matching groups' rules MUST be combined into one group

Extensions to the robots.txt are mentioned [section 2.2.4 "Other Records"](https://www.rfc-editor.org/rfc/rfc9309.html#section-2.2.4) with the requirement that it does not terminate groups:
> Parsing of other records MUST NOT interfere with the parsing of explicitly defined records in [Section 2](https://www.rfc-editor.org/rfc/rfc9309.html#specification). For example, a "Sitemaps" record MUST NOT terminate a group.

Because RFC 9309 does not define extensions, it does not tell anything about their scope and how extensions from multiple rule groups are to be merged. Nevertheless, group merging should obviously also apply for extensions, e.g.

```
User-agent: ExampleBot
Crawl-delay: 10
allow: /

user-agent: *
disallow: /search
disallow: /login/

user-agent: ExampleBot
Clean-param: sessionid
disallow: /login/
```

Note: the tests are affected by #590 and need to be updated after rebase.

### DIFF
--- a/src/test/java/crawlercommons/robots/SimpleRobotRulesParserTest.java
+++ b/src/test/java/crawlercommons/robots/SimpleRobotRulesParserTest.java
@@ -1495,6 +1495,82 @@ public class SimpleRobotRulesParserTest {
         assertEquals("10/1m", data.getValues().get(0));
     }
 
+    /**
+     * Test the combination of merging rule groups and the Crawl-delay extension.
+     *
+     * @throws Exception
+     */
+    @Test
+    void testCrawlDelayGroupMerging() throws Exception {
+        byte[] robotstxt = readFile("/robots/crawl-delay-group-merging.txt");
+
+        SimpleRobotRulesParser robotParser = new SimpleRobotRulesParser();
+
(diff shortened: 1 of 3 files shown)


sebastian-nagel:
Needs also be rebased on #595, which makes the detection of the `Content-Signal` directive fail.

jnioche (approved):
LGTM

sebastian-nagel:
- Rebased on master after #590 and #595 have been merged.
- Uncommented the two assertions which were failing because of the `visit-time` leaked from the wildcard rule group.

PR #597, opened 2026-07-27
Fix BasicURLNormalizer rejecting URLs with an IPv6 address as host name (#587)

### LINKED ISSUE #587: BasicURLNormalizer rejects URLs with IPv6 address as host name
If a URL has an IPv6 address as host name, BasicURLNormalizer rejects it as invalid, returning null:
```
$> echo "http://[::1]/" | java -cp ... crawlercommons.filters.basic.BasicURLNormalizer
[main] INFO crawlercommons.filters.basic.BasicURLNormalizer - http://[::1]/ => null
```

The issue is caused by
```java
        // escape to ensure URL does not contain illegal characters
        urlString = escapePath(urlString);
```
which converts the URL erroneously to `http://%5B::1%5D/`.

Discovered while revieing #586. Thanks, @rzo1!

rzo1:
Fixes #587.

filter() escaped the whole URL string via escapePath(), which percent-encodes [ and ]. An IPv6 host was mangled into http://%5B::1%5D/ and subsequently rejected as malformed.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,7 @@
 Crawler-Commons Change Log
 
 Current Development 1.7-SNAPSHOT (yyyy-mm-dd)
+- BasicURLNormalizer rejects URLs with IPv6 address as host name (Richard Zowalla, sebastian-nagel) #587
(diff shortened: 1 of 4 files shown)


jnioche (approved):
LGTM

PR #602, opened 2026-09-09
Publish a daily SNAPSHOT to Maven Central

### LINKED ISSUE #601: Publish nightly snapshot
We release infrequently and it would be nice to users to be able to test new features early on.

jnioche:
Closes #601

Adds a scheduled workflow deploying the current SNAPSHOT to the [Central Portal snapshot repository](https://central.sonatype.com/repository/maven-snapshots/), so that users can test new features between releases.

Snapshots are not validated by Central, so the build skips signing, javadoc and sources. Tests are skipped too, as they already run on every push to master via the build workflow.

Deployment is skipped when the POM version is not a snapshot. The scheduled run is additionally skipped when nothing was committed since the previous one, to avoid republishing an identical snapshot every day; a manual `workflow_dispatch` run always deploys.

Notes:
- reuses the `SONATYPE_USERNAME` / `SONATYPE_PASSWORD` secrets, no GPG secret needed
- `-P release` is required as `central-publishing-maven-plugin` is declared in that profile
- SNAPSHOT publishing needs to be enabled for the `com.github.crawler-commons` namespace on the Central Portal


jnioche:
Will merge - it does not impact anything else

PR #604, opened 2026-09-10
[Robots.txt] Scope the Content-Signal extension to the user-agent group

jnioche:
Declare `RobotsExtension.CONTENT_SIGNALS` as a per-group directive instead of a global one, before 1.7 is released.

## Problem

Content Signals ([contentsignals.org](https://contentsignals.org/), [Cloudflare](https://blog.cloudflare.com/content-signals-policy)) are written *inside* a user-agent group, exactly like `Disallow` and `Crawl-delay`:

```
Content-Signal: ai-train=maybe

User-agent: *
Content-Signal: ai-train=no
Disallow:

User-agent: mybot
Content-Signal: ai-train=yes
Disallow:
```

`CONTENT_SIGNALS` was added in #569 with `perGroup=false`, so every occurrence in the file is collected into one flat list with no record of the group it came from. On the file above, parsing for `mybot` returns:

```
CONTENT_SIGNALS -> [ai-train=maybe, ai-train=no, ai-train=yes]
```

A consumer cannot tell which value applies to it. Any reduction it picks — first, last, or a merge — is wrong for some input, and the failure is silent: a site whose `*` group says `ai-train=no` while an unrelated group says `ai-train=yes` yields a list in which the permissive value is indistinguishable from the applicable one. For a directive whose purpose is expressing an AI-training opt-out, resolving it to the wrong group is the one outcome worth designing against.

It also makes the directive unusable for the question it most obviously serves — "what does this site tell *me*?" — which today still requires hand-parsing robots.txt alongside the library.

## Change

One flag:

```diff
-    CONTENT_SIGNALS("content-signal", false, "content-signals"),
+    CONTENT_SIGNALS("content-signal", true, "content-signals"),
```

No other production code is needed. `handleExtensionDirective()` already branches on the flag and routes per-group values through the group-matching state machine, and `ParseState.clearPerGroupExtensions()` (#590) already discards wildcard-group values once a specific group matches. So the directive inherits RFC 9309 semantics for free: wildcard fallback, specific-group override, and merging across several matching groups. With the change, the same file gives:

| target agent | `CONTENT_SIGNALS` |
|---|---|
| `mybot` (own group) | `[ai-train=yes]` |
| `somebot` (no group → `*`) | `[ai-train=no]` |

The rest of the diff is the javadoc, the changelog, and tests.

## Scope

`RobotsExtensionData` deliberately holds raw strings and this PR keeps that boundary: splitting `search=yes, ai-train=no` into pairs, lower-casing, and handling the optional path-prefix form (`Content-Signal: /images/ ai-train=no`) stays with the consumer. Only the *scoping* is corrected here.

## Tests

New `content-signal-groups.txt` fixture plus tests covering: specific group wins over `*`; fallback to `*` for an agent without its own group; merging of two groups both matching the agent; values of a non-matching group never appearing; the `content-signals` alias scoped identically.

`extensions-group-merging.txt` and `extension-robots.txt` gained per-group `Content-Signal` lines so the existing fixtures exercise the new scoping, and the affected assertions in `testExtensionsGroupMerging` / `testEnableAllExtensions` / `testEnableAllExtensionsWildcardAgent` were updated. Full suite is green.

## Points for review

- **Timing.** `RobotsExtension` is public API but 1.7 is unreleased — the enum exists only in `1.7-SNAPSHOT`, so flipping the flag now costs nothing. After 1.7 ships this becomes a silent behavioural change for anyone who built on the global reading, and would need a deprecation path or a second enum member.
- **Directive before the first group.** A `Content-Signal` line above any `User-agent` line is currently captured; with `perGroup=true` it is dropped, since `isAddingRules()` is false. This is a real behaviour change, and it is a consequence of how per-group extensions are scoped rather than something the Content Signals specification settles — @sebastian-nagel rightly pushed back on the earlier version of this bullet, which justified the drop by analogy with `Disallow`: RFC 9309 explicitly requires a rule block to start with a `User-agent` line, so that analogy does not carry over to an extension the RFC says nothing about. There are also sites that write the directive at the top of the file with an obviously global intention. `extensions-group-merging.txt` had exactly such a line, annotated "global scope"; it is kept in the fixture and now asserted to be dropped, and the reasoning is recorded on `testContentSignalBeforeFirstGroup`. Collecting out-of-group values separately, so that users can decide how to use them, is left open as a follow-up.
- **Consistency rule for the flag.** `HOST` and `LLM_POLICY` stay global, which looks right for both — they are file-level statements about the site rather than per-crawler instructions. If you want a written rule for deciding the flag, the one this PR follows is: *does the directive answer "what does this site tell **this crawler**?" (per-group) or "what is true of this site?" (global)*.
- I could not find a normative statement of the scoping on contentsignals.org to cite; the argument here rests on the syntax and on Cloudflare's generated files.


src/main/java/crawlercommons/robots/RobotsExtension.java
```diff
@@ -72,15 +72,34 @@ public enum RobotsExtension {
      * Cloudflare Content Signals policy directive, an extension for content
      * governance rules (e.g. AI training data usage, attribution). See <a
      * href="https://blog.cloudflare.com/content-signals-policy">Cloudflare blog
-     * post</a>.
+     * post</a> and <a href="https://contentsignals.org/">contentsignals.org</a>.
```
sebastian-nagel, line 75: +1 to add the second source.

src/main/java/crawlercommons/robots/RobotsExtension.java
```diff
@@ -72,15 +72,34 @@ public enum RobotsExtension {
...
+     * 
+     * <p>
+     * The directive is written inside a user-agent group and is therefore
```
sebastian-nagel: This sounds overtly detailed here. Maybe add this to the description of `RobotsExtension.isPerGroup()` and link to it.

src/main/java/crawlercommons/robots/RobotsExtension.java
```diff
@@ -72,15 +72,34 @@ public enum RobotsExtension {
...
+     * User-agent: *
      * Content-Signal: search=yes, ai-train=no
+     * Disallow:
+     * 
+     * User-agent: mybot
+     * Content-Signal: search=yes, ai-train=yes
+     * Disallow:
      * </pre>
+     * 
+     * <p>
```
sebastian-nagel, line 95: Again: this is the case for all robots extensions. Could link to `RobotsExtensionData`.

src/test/java/crawlercommons/robots/SimpleRobotRulesParserTest.java
```diff
@@ -1571,6 +1573,92 @@ void testExtensionsGroupMerging() throws Exception {
...
+        // an agent without a group of its own falls back to the wildcard group
+        SimpleRobotRules rules = robotParser.parseContent(FAKE_ROBOTS_URL, robotstxt, "text/plain", Set.of("somebot"));
+        RobotsExtensionData data = rules.getExtensionData(RobotsExtension.CONTENT_SIGNALS);
+        assertNotNull(data);
+        assertEquals(List.of("search=yes,ai-train=no"), data.getValues());
+    }
+
+    /**
+     * A Content-Signal line above the first User-agent line does not belong to
+     * any group and is dropped, in the same way as a Disallow rule outside of
```
sebastian-nagel: I had to think longer about this:

1. I've seen quite a few robots.txt files which do it differently and it's clear that outside a group means global scope. See the robots.txt of [defensestorm.com](https://defensestorm.com/robots.txt) or [ensmart.ai](https://ensmart.ai/robots.txt). For others ([awsociety.ly](https://lawsociety.ly/robots.txt)), the content signals are in a group following the grouping rules of RFC 9309, but it's clear that the intended scope is global.
2. The content signals are sparsely specified. From "You can apply rules to specific bots instead of all of them.", one could also assume, that no scope means global scope.

In this sense, the reference to the "Disallow" directive might be misleading. RFC 9309 is clear and states that a rule block has to start with a "User-agent" line.

sebastian-nagel (approved):
Good catch, @jnioche!

The "Content-Signal" directive is surely meant to be scoped to a user-agent group. Although I'm a little bit unsure how we should treat appearances outside a group. Maybe collect them separately, so that users can decide how to use them. But that would be another issue.

jnioche:
thanks for the review @sebastian-nagel, your suggestions should have been addressed. 
Would you mind having a quick look before merging? Thanks

sebastian-nagel (approved):
Thanks, @jnioche!

